PR #90, opened 2016-08-25
Provide 8/25 status update

cgreene:
Incorporate a discussion of work to date and provide a set of steps to get new participants up to speed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,5 +55,6 @@ want to take to get up to speed quickly.
    review.
 
 In about a week, we plan to move into the phase where we start to vigorously
-discuss this question for each area that the review will cover (currently also
-being discussed in [#88](https://github.com/greenelab/deep-review/issues/88)).
+argue about the answer to the question that we coalesce on with
+[#88](https://github.com/greenelab/deep-review/issues/88) for each area that the
+review will cover.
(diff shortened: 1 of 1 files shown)


cgreene:
By the way - codeclimate is angry that the list elements 2, 3, and 4 don't start at '1'. I'm not sure why the markdown linter is flagging that.

README.md
```diff
@@ -32,3 +32,28 @@ This repository provides a home for the paper. We'll operate on a pull request
...
+   of the paper.
+3. Choose some papers in an area that you care about, review them, and summarize
+   them.
+4. Dive in to participation in
+   [#88](https://github.com/greenelab/deep-review/issues/88) and help us to
+   further refine the specific question that we're going to deal with in this
+   review.
+
+In about a week, we plan to move into the phase where we start to vigorously
+discuss this question for each area that the review will cover (currently also
```
gwaygenomics: Maybe instead of "discuss this question" it would be more clear to state something like: "discuss the overarching question" or "discuss the main question" or "discuss the guiding question".

gwaygenomics:
I looked into the markdownlint question a bit - looks like we would have to set [Rule MD029](https://github.com/mivok/markdownlint/blob/master/docs/RULES.md) to `ordered` for this check to pass. I don't know why ordered lists and unordered lists are mutually exclusive to the program - but it appears to be a non-issue. 

LGTM

agitter:
It's already merged, but I like the status update

PR #106, opened 2016-10-13
Add in stub files so that we can start writing.

cgreene:
**Motivation**: The organization of the repository is unclear. This adds stub files and a pandoc command to build a PDF.

**Changes**: We've added a sections directory, numbered sections, and the bash script that calls pandoc.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,2 @@
+
+pandoc -f markdown -o review.pdf sections/*.md
--- a/build.sh
+++ b/build.sh
@@ -1,2 +1,2 @@
 
-pandoc -f markdown -o review.pdf sections/*.md
+pandoc --verbose --from=markdown --output=review.pdf sections/*.md
--- a/sections/00_title_authors.md
+++ b/sections/00_title_authors.md
@@ -1,0 +1,5 @@
+% Deep Learning, Genomics, and Precision Medicine
+% Casey Greene;
+  Anthony Gitter;
+  More TBD;
+%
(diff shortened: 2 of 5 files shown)


agitter:
What review process do you prefer for pull requests?  Does it matter that Code Climate found 4 issues even though they appear to be reasonable formatting deviations  I haven't used GitHub's new Pull Request Review feature yet but am willing to try it out.

cgreene:
@agitter : we usually use codeclimate to provide a quick check - not a hard and fast set of rules. The h1 issue crops up because codeclimate doesn't realize that we're going to merge the markdown files. The issue of punctuation on the header is a bit stylistic.

I would disable those two in a config since they will keep cropping up, but it doesn't seem to be possible: https://docs.codeclimate.com/docs/markdownlint#configure-the-engine

I guess I'd say, we should accept PRs that we could stand by. If we think there are issues, we should point them out and discuss them. In this case, I guess we are more or less the primary maintainers of the repo so the buck probably has to stop with us.

cgreene:
In terms of process - either a ":+1: Looks good to me" or a full review (using the new github feature) is fine depending on the size of the commit. We tend to use both in our group.

cgreene:
@agitter : agree on figuring out who is actually going to write. Any ideas for the best way to do that?

agitter:
@cgreene Maybe we can look at who has contributed to issues and discussion once more and give them last call invites to help write if they haven't already explicitly declined?

agitter:
@cgreene I created #116 to poll everyone and figure out who wants to write with us.  You and I can continue working on stubs and then help map writers to sections.

cgreene:
@agitter Thanks! Had issues getting to github throughout the day (sounds like a DNS attack happened) so I am behind on my deep review items that I had hoped to get to today!

PR #108, opened 2016-10-13
Start to stub in categorize section

cgreene:
Includes chunks that someone can bite off if they want to focus on a specific area for writing.

### DIFF
--- a/sections/02_intro.md
+++ b/sections/02_intro.md
@@ -3,14 +3,14 @@
 ### Potential writing prompt
 
 One potential future that we could imagine is a world in which data, once
-gathered, is rapidly integrated into our understanding of living systems
-(STUDY). What we learn is rapidly put to use. Our health-related activities are
-constantly monitored (e.g. by wearables) and all of our interactions with health
-care systems are extensively tracked. These sources of information are combined
-to help to guide our health care and maintenance. (CATEGORIZE) We'd be able to
-compare our state and trajectory to (anonymized) others, and identify means to
-improve our health. (TREAT) These means might contain drug combinations selected
-based on personalized predictions.
+gathered, is rapidly integrated into our understanding of living systems. What
+we learn is rapidly put to use. Our health-related activities are constantly
+monitored (e.g. by wearables) and all of our interactions with health care
+systems are extensively tracked. These sources of information are combined to
+help to guide our health care and maintenance. We'd be able to compare our state
+and trajectory to (anonymized) others, and identify means to improve our health.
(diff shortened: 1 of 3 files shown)


cgreene:
@agitter : totally agree that italics for prompts and suggestions would be most helpful. I'll convert to that (and fix a couple of the markdown issues that CC noted). Thanks!

agitter:
I added a placeholder for background information on deep learning, which came up in #116.

cgreene:
@agitter : ok - tried to address. Some of the codeclimate issues will take care of themselves as we write and clarify our thoughts (e.g. header length/punctuation). Basically these are all stubs to be replaced. Thanks!

sections/02_intro.md
```diff
@@ -1,3 +1,25 @@
...
+care systems are extensively tracked. These sources of information are combined
+to help to guide our health care and maintenance. (CATEGORIZE) We'd be able to
+compare our state and trajectory to (anonymized) others, and identify means to
+improve our health. (TREAT) These means might contain drug combinations selected
+based on personalized predictions.
+
+### If this happens, is deep learning required for any of it? Are we any closer
+### because of the advent of deep learning?
+
+* "Categorize" and "treat" sound a bit like PMI goals. Another way to think
```
gwaygenomics: I don't think you meant to bullet this?

gwaygenomics: also, can you expand the acronym for PMI the first time you reference it?

sections/02_intro.md
```diff
@@ -1,3 +1,25 @@
 ## Introduction
 
-Intro goes here.
+### Potential writing prompt
+
+One potential future that we could imagine is a world in which data, once
```
gwaygenomics: I definitely agree with the three subcategories and the content of this stub. 

But, for consistency, can you mention them either before or after the period? Right now I am not sure which sentences `(CATEGORIZE)` or `(TREAT)` are referencing

agitter:
My comments have been addressed

sections/02_intro.md
cgreene: @gwaygenomics : removed them for now. We'll see how the sections come along to decide what to talk about specifically.

sections/02_intro.md
cgreene: Bullets here (and elsewhere) seem to be induced by atom's convenient "reflow selection" functionality. I tracked it down here and in the other part, and expanded PMI. Thanks!

cgreene:
@gwaygenomics : let me know if I've addressed all of your comments! thanks!

PR #115, opened 2016-10-20
reality has struck!

cgreene:
Don't think Nov. 1 is realistic. Dec. 1 still gives us a bit of time to convert markdown to the format that the journal requires by the journal deadline.


agitter:
Agreed, we're not going to have a draft in ~10 days.

PR #118, opened 2016-10-24
Discussion stub

agitter:
I added ideas for general sections that span the Categorize, Study, and Treat subject areas.  I also created a Discussion section separate from the Conclusions.

### DIFF
--- a/sections/06_discussion.md
+++ b/sections/06_discussion.md
@@ -1,3 +1,54 @@
 ## Discussion
 
-*Discussion here*
+*This section provides meta-commentary that spans the Categorize, Study, and
+Treat subject areas.  The candidate sub-sections below are initial ideas that
+can be further pruned.*
+
+### Evaluation
+
+*What are the challenges in evaluating deep learning models that are specific to
+this domain?  This can include a discussion of ROC versus precision-recall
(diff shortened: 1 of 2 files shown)


sections/06_discussion.md
```diff
@@ -0,0 +1,54 @@
...
+to the average computational group, and if so, will that limit future progress?
+For instance, some hyperparameter search strategies are not feasible for a lab
+with only a couple GPUs.*
+
+*Some of this is also outlined in the Categorize section.  We can decide where
+it best fits.*
+
+### Code, data, and model sharing
+
+*Anything to be said?*
```
gwaygenomics: I think it would be useful to discuss reproducibility here - even though the importance of reproducibility is not specific for our topic. 

However, when deep models are being built the hyperparameter space is so large and data is so heterogeneous we may need to think about how reproducibility in deep learning in biomedical science, and the impact it has on improving healthcare, is different than reproducibility in the broader computational biology context. To slightly rephrase this question: "what might a clinician, or a policy maker, need to see in order to influence treatment decisions". 

Might this be a similar situation as a requirement for reproducible GWAS signals?

cgreene:
@agitter : feel free to merge at will!

PR #120, opened 2016-10-26
2016-10-26 status report

agitter:
As suggested in #116, all co-authors can help outline the review with pull requests.  That will organize our discussion of specific sub-sections.  I'm not sure how to tag everyone who is interested in a particular topic with relevant pull requests.  Perhaps we all will need to watch the repository to see when relevant pull requests are made.


agitter:
I reordered the status updates so that the most recent comes first.

Issue #88, opened 2016-08-22
Refine our guiding question.


cgreene:
Throughout our initial review of papers, our guiding question has been: _What would need to be true for deep learning to transform how we categorize, study, and treat individuals to maintain or restore health?_

I think we should take a bit of time to start discussing this. I expect that we will probably want to refine it (e.g. what does 'transform' even mean in this context). We should bring in examples where pertinent - for example I might reference #55 if I wanted to discuss the role that new in silico drug development strategies could play in this transformation.

cgreene:
@agitter @gwaygenomics @brettbj @michaelmhoffman @sw1 @akundaje
Trying to tag those who have participated to fully review at least one paper thus far to make sure that we get a chance to participate in the refinement phase.

agitter:
A lower bar than 'transform' would be to ask what it would take to have the practitioners who categorizes, study, and treat individuals use or benefit from deep learning approaches.  The answer could involve the quality of predictions as well as the accessibility and interpretability of the models.

Working backwards, we can also think about what questions are useful for organizing the papers we want to discuss.  As @gwaygenomics said (#87) we want to avoid reproducing existing reviews, which could be harder if we group papers by 'imaging', 'EHR', 'genomics', etc.  Two immediate thoughts:

- **Stratify papers based on what neural networks contribute to the problem.**  Category 1 might include what @sw1 called straightforward ML papers (#37) where the neural network provides non-linear feature transformations and not much else.  Intermediate groups could include papers that use elements like multi-task learning (#13, #16, #55, etc.).  Other attributes would include convolutions in 1, 2, or 3 dimensions.  Downsides are that this would focus more on the machine learning than the applications and could lead to jumping around from application to application.
- **Group papers by how directly relevant they are to treating individuals.**  The gradient could be:
    - Basic molecular biology (e.g. #49)
    - Disease study but more on the molecular biology or genetics side than clinical (e.g. #79)
    - Clinical study about phenotypes in general, not individuals (e.g. #78)
    - Relevant to individuals' disease risk, health, or treatment (e.g. #5, #81)

hussius:
[...]

PR #121, opened 2016-10-27
Update 02_intro.md

traversc:
Proposed paragraph in introduction exploring type of biological problems people have looked at in the literature.

### DIFF
--- a/sections/02_intro.md
+++ b/sections/02_intro.md
@@ -23,3 +23,11 @@ for the precision medicine initiative (PMI)?*
 
 *Definitions, specific architectures, etc.  We may want to clarify what we mean
 by "deep" learning when most methods use few hidden layers.*
+
+### Types of biological problems
+
+This section could contain the types of biological problems people have already applied deep learning to, based on the existing literature.  This could be mentioned in two ways: 
+
+1. The type of input data: DNA sequence, protein sequence, clinical data, imaging data, etc.
+2. The output/objective of the model: protein structure/function, protein localization, miRNA/TF target prediction, patient diagnosis/prognosis, drug discovery/prediction, or unsupervised clustering, etc.
+


agitter:
I generally support some form of introduction to the biomedical problems and data types that will be covered in the review.  Most of those will fall neatly into one of Study/Categorize/Treat and will be discussed more extensively in those sections.  Here, we can have a fairly short preview of what is to come.

I think we want to avoid enumerating methods in the style of "neural network architectures A, B, C have been applied to data types X, Y, Z" in both the intro and main sections.  That has been well-covered by existing reviews, and we hope to target this one more toward the guiding question of #88 and provide perspective.

I'll leave this open for a couple days to see if there are any comments from others.

cgreene:
I agree with @agitter that we want to avoid enumerating either domains or architectures because there are reviews that already exist of that form and we were specifically asked not to do that by the journal:

> We intend the reviews to address critical developments in an area of cross-disciplinary research and, when possible, to place such research in a broader context. This is not a place for comprehensive literature surveys.

The additional language on the readme highlights an area that I think we can contribute - namely informed speculation.

> We do encourage you to speculate in an informed way, and to be topical and provocative about the subject without worrying unduly about space, (the provisional target length is 8 -12,000 words). Please think of this as an article which will be a landmark in your area, and will come to be considered as a classic paper of the literature.

cgreene:
I do wonder if this would be better placed in the "study" section. I think we may want to discuss specific approaches largely within those to make the introduction and conclusion more speculative than specific-case driven (unless there's one or two that we really want to highlight). There is a bit of this discussion of application areas in our 'categorize' outline.

agitter:
Regarding architectures, I think we need to say enough so that someone who isn't familiar with neural networks can appreciate why they are special (worthy of a review), why a CNN is different from a RNN, etc., but no more.

Regarding domains, it may be helpful to give a few examples in the intro to help reader understand what we mean by Categorize, Study, Treat.  The individual sections could expand on that.

agitter:
Since there is no more discussion, I'll merge this with the caveats raised above:
* We don't want to enumerate domains or architectures
* A short, targeted discussion of the variety of problems will we present could help give our audience a preview of the scope of this review.  But most of the discussion of different application areas will fall within the Categorize, Study, Treat sections.

Issue #2, opened 2016-08-03
Overall Manuscript Structure


cgreene:
The overall aims of the Headline Review articles are outlined in the README. Here's a document structure that I am playing around with to target the review at this question: *What would need to be true for deep learning to transform how we categorize, study, and treat individuals to maintain or restore health?*

* Relevant areas where methods inspired by deep learning are already having an impact
    * Sequence -> Function
    * Transcriptional regulation
    * Patient information
    * Imaging + Bio
* The structures of problem statements which use deep learning towards these ends
    * Supervised approaches
        * Convolutional NNs on genome
        * Principles in which multiple synergistic patterns are learned simultaneously
        * more examples of shared properties across approaches
    * Unsupervised approaches
        * some denoising autoencoder work is common across systems
        * more shared properties
* Perspectives towards the future & overall question.
    * Which challenges do we think will be resolved first?
    * Are there any approaches/data types that have taken off in other fields but that are under-utilized here?
    * What initiatives or data do we think are particularly interesting for/amenable to deep learning analyses and why?
* Overall summary on state of the field & reflection towards overall question.

There are some wonderful github-based reading groups/lists by @pimentel @hussius @gokceneraslan. If any of you have feedback as we structure this review, please provide it. If you'd like to participate - dive in!

cgreene:
Also wanted to tag @YosephBarash and @davek44 who have been active in this area for their thoughts.

michaelmhoffman:
What is the format of the text itself? GitHub markdown?

michaelmhoffman:
Does this need to actually have anything to do with "systems pharmacology"?

cgreene:
@michaelmhoffman It should probably roughly touch on topics that could be construed as systems pharmacology. My read is that the precision medicine perspective + deep learning on genomic/transcriptomic/proteomic/etc data gets us close enough.

Format of the text itself will be markdown [eventually I'll convert it to LaTeX and reformat]. I think we will use something like [@doi:<doi link>] for citations. @dhimmel has code to automatically pull down doi metadata and covert to bibtex.

agitter:
> Are there any approaches/data types that have taken off in other fields but that are under-utilized here?

Reinforcement learning perhaps?  http://karpathy.github.io/2016/05/31/rl/ gives a brief intro.  I'm not aware of examples in biology or medicine.

michaelmhoffman:
Reinforcement learning hooked up to some experimental system would be fun.

[...]

PR #123, opened 2016-10-29
Enable tag citations

dhimmel:
See https://github.com/greenelab/deep-review/issues/2#issuecomment-256496372.

@agitter please review!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,22 +35,29 @@ additional feedback.
 
 ## Markdown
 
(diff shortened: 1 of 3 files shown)


dhimmel:
@cgreene satisfying code climate was a burden (and of limited utility IMO).

cgreene:
@dhimmel : you don't have to satisfy code climate to merge. It's an indicator, not a rule.

agitter:
Thanks, this is perfect.

PR #134, opened 2016-11-05
Treat stubs

agitter:
@kumardeep27 This is one of your main areas of interest, so let's work together to outline this section.  Are there major papers or topics I have missed?

@cgreene There is a fuzzy boundary between Categorize and Treat papers in some cases.  I proposed that papers that categorize patients for some specific clinical decision be covered here.  What do you think?

### DIFF
--- a/sections/05_treat.md
+++ b/sections/05_treat.md
@@ -1,0 +1,38 @@
+## The impact of deep learning in treating disease and developing new
+treatments
+
+*There will be some overlap with the Categorize section, and we may have to
+determine which methods categorize individuals and which more directly match
+patients with treatments.  The sub-section titles are merely placeholders.*
+
+### Categorizing patients for clinical decision making
+
+*How can deep learning match patients with clinical trails, therapies, or
+other interventions?  As an example, [@doi:10.1016/j.jalz.2015.01.010]
+predicts individuals who are most likely to decline during a clinical trial
+and benefit from the treatment.*
+
+### Effects of drugs on transcriptomic responses
+
+*We discussed a few papers that operate on Library of Network-Based Cellular
+Signatures (LINCS) gene expression data.  We could briefly introduce the
+goals of that resource and comment on the deep learning applications.  In the
+Issues, we had reservations about whether the improvements in expression
+prediction are good enough to make a practical difference in the domain and
+feature selection and construction.*
+
+### High-throughput screening and drug development
+
+*Deep learning has been applied in some fashion to structure-based,
+compound-protein interaction-based, and ligand-based prediction problems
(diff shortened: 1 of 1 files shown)


cgreene:
I agree that there is significant overlap. Technically, we *have* to
categorize somebody to treat them. I guess we could reframe this to
"develop treatments." I would also be fine with specific treatment
decisions going here, and more discovery-oriented categorization be in the
"categorize" section. I guess that what I'm saying is - I agree that there
is an issue, and either potential solution seems fine to me! :+1:

agitter:
@cgreene From a practical standpoint, I think we have more to discuss in the Categorize section so leaving it here would make the three main topics somewhat more balanced.  I'm also okay moving it to Categorize once we start writing if we see that it better aligns with the rest of that section.

cgreene:
Sounds good!

agitter:
@kumardeep27 Last chance for comments on the structure before I merge.  We only have a couple weeks to write so we should finalize the outline for this section and get started.

XieConnect:
@agitter  @cgreene  I agree with your comments on overlaps, and am fine with either decision. 
I personally think it is better to merge them into one giant section about deep learning on EHR or clinical documents. There are several reasons: 

First, we may not have enough space to cover these two closely related topics separately. 

Second, these two tasks are closely related in terms of data used and methods applied (e.g., classification problems). And the differences seem more on the semantic level (do you treat expert manual labels as true labels -- in "Categorize", or do you treat future holdout set of disease development as true labels -- in "Treat"). 

Third, I don't think there are many (if any) success stories in clinical deployment of machine learning-based decision support methods in general for clinical settings, let alone deep learning. I may need to read those works mentioned by @agitter  though. 

- Wei

cgreene:
I am going to merge this for now. We can revisit @XieConnect's point as the sections start to take shape, to see if the sections are unbalanced and should be merged.

We should not be constrained with regards to length. As I understand it, this specific invitation allows us 8000-12000 words. If we absolutely need more than 12000 words to write the review that we think is needed, I am under the impression that we can ask.

XieConnect:
Sure. Sounds great. 

- Wei

agitter:
> Third, I don't think there are many (if any) success stories in clinical deployment of machine learning-based decision support methods in general for clinical settings, let alone deep learning.

@XieConnect In this manuscript, we have some freedom to discuss not only the current state of the art but also where we think the field is moving or should be moving.  We can present existing successes (or failures) as well as deep learning applications with immense future potential.

XieConnect:
@agitter  That'll be great and much more valuable than a standard systematic survey! I'll try to comment on specific works included in this direction later.

PR #165, opened 2016-12-21
12/21 Status Report

cgreene:
Purpose: To provide visitors to the repository with an updated status message.

Changes: Note that we are now actively writing the manuscript. Move the date to Jan 15, 2017.


cgreene:
added links to #147 and @evancofer to provide an example. Merging now.

PR #167, opened 2016-12-22
Pattern Discovery / Categorize

cgreene:
Motivation: We need to start filling in the sections with relevant literature.

Progress: This includes some thoughts on unsupervised EHR analysis and some extraction work via NLP. I have stubbed in portions for @sw1 and potentially @traversc. If you guys want to contribute you can either work of of this PR and make a pull request into this branch of my repo, or you can start writing now and create a PR against the repo once this goes through. My impression is that @sw1 is the lead on this portion given his previous volunteering.

Timing: I estimate that I'll get to wrap up the text around EHRs tomorrow. I just got a paper that I was struggling to get that seems relevant. I need some time to review it. I will also aim to talk some about the imaging work, but that may be a separate PR.


cgreene:
No reviewers directly named at this time - will name reviewers after I get to make some changes tomorrow.

sections/03_categorize.md
```diff
@@ -28,7 +28,76 @@ we share these data while respecting privacy? I've made a section for each of
...
+features for domain-specific concept extraction [@arxiv:1611.08373]. They found
+that performance was in line with, but did not exceed, existing state of the art
+methods. The deep learning method had performance lower than the best performing
+domain-specific method in their evaluation[@arxiv:1611.08373]. This highlights
+the challenge of predicting the eventual impact of deep learning on the field.
+This provides support that deep learning may impact the field by reducing the
+researcher time and cost required to develop specific solutions, but it may not
+lead to performance increases.
+
+TODO: survival analysis/readmission prediction methods from EHR/EMR style data
```
sw1: Nevertheless, recent work has revealed one domain in which deep networks have proven superior to traditional methods. Survival analysis models the time leading to an event of interest from a shared starting point, and in the context of EHR data, often associates these events to subject covariates. Exploring this relationship is difficult, however, given that EHR data types are often heterogeneous, covariates are often missing, and conventional approaches require the covariate-event relationship be linear and aligned to a specific starting point (@arxiv:1608.02158). Early approaches, such as the Faraggi-Simon feed-forward network, aimed to relax the linearity assumption, but performance gains were lacking (@doi:10.1016/S0167-9473(99)00098-5). Katzman et al. in turn developed a deep implementation of the Faraggi-Simon network that, in addition to outperforming Cox regression, was capable of comparing the risk between a given pair of treatments, thus potentially acting as recommender system (@arxiv:1606.00931). To overcome the remaining difficulties, researchers have turned to deep exponential families, a class of latent generative models that are constructed from any type of exponential family distributions (@arxiv:1411.2581v1). The result was a deep survival analysis model capable of overcoming challenges posed by missing data and heterogeneous data types, while uncovering nonlinear relationships between covariates and failure time. They showed their model more accurately stratified patients as a function of disease risk score compared the current clinical implementation.

traversc: This sounds good, I have nothing more to add, except for possibly mentioning other non-linear methods.  I can think of Random Forest used in survival analysis (@arxivL0811.1645)

cgreene:
Ok - given that @sw1 has some content to contribute this is probably a good time to remove the [WIP] tag and aim to merge. This is a bit more of a laundry list than I like, but once we get the pieces in place we should return for some deeper analysis. This has some placeholders for expected contributions.

agitter:
Acknowledging the review request, I should be able to do it within 24 hours

sections/06_discussion.md
```diff
@@ -137,3 +137,7 @@ interpretability problems, how can we best ensure reproducible models? What
 might a clinician, or policy maker, need to see in a deep model in order to
 influence healthcare decisions? Or, is deep learning a hypothesis generation
 machine that requires manual validation?*
+
+### Transfer learning/transferability of features
```
agitter: As you point out, transfer learning and multi-task learning has come up as a theme in many papers and domains.  Do you suggest we defer most of that discussion in the domain-specific sections and coalesce it here?  Or introduce it in the specific sections and have a cross-domain recap here?

sections/03_categorize.md
```diff
@@ -19,7 +19,7 @@ tackle any of them? Are there example approaches whereby deep learning is
 already having a transformational impact? I (Casey) have added some sections
 below where I think we could contribute to the field with our discussion.*
 
-### Major challenges
+### Major Areas of Existing Contributions
```
agitter: Should we settle on title case or lower case for section and sub-section headings?

sections/03_categorize.md
```diff
@@ -28,7 +28,148 @@ we share these data while respecting privacy? I've made a section for each of
...
+
+One of the general areas where deep learning methods have had substantial
+success has been in image analysis. Applications in areas of medicine that use
+imaging extensively are also emerging. Mammography has been one area with
+numerous contributions [@doi:10.1007/978-3-319-46723-8_13,
+@doi:10.1007/978-3-319-24553-9_74, @doi:10.1101/095794, @doi:10.1101/095786]. In
+all of this work, the researchers must work around a specific challenge - the
+limited number of well annotated training images. To expand the number and
+diversity of images, the researchers have employed approaches where they employ
+adversarial examples [@doi:10.1101/095786] or first train towards human-created
```
agitter: If we haven't introduced adversarial networks earlier in the manuscript, it may deserve more attention.

sections/03_categorize.md
```diff
@@ -28,7 +28,148 @@ we share these data while respecting privacy? I've made a section for each of
...
+features before subsequent fine tuning [@doi:10.1007/978-3-319-46723-8_13]. The
```
agitter: What does "train towards human-created features" mean?  It sounds interesting.  Is this initializing weights using prior knowledge?  Training a network with manually-defined features first and then using those weights to initialize a different network?

sections/03_categorize.md
```diff
@@ -28,7 +28,148 @@ we share these data while respecting privacy? I've made a section for each of
...
+detectors. Then the small number of labeled examples could be used for
+subsequent training. Similar strategies have been employed for EHR data where
+high-quality labeled examples are also difficult to obtain
+[@doi:10.1101/039800].
+
+In addition to radiographic images, histology slides are also being analyzed
+with deep learning approaches. In recent work, Wang et al.[@arxiv:1606.05718]
+analyzed stained slides to identify cancers within slides of lymph node slices.
+The approach provided a probability map for each slide. On this task a
+pathologist has about a 3% error rate. Their algorithm had about a 7% error
```
agitter: Do we need to be more precise or is this the terminology used in the paper?  Does their error rate include false positives and false negatives?

sections/03_categorize.md
```diff
@@ -28,7 +28,148 @@ we share these data while respecting privacy? I've made a section for each of
...
+
+In recent work, Yoon et al.[@doi:10.1007/978-3-319-47898-2_21] analyzed simple
+features using deep neural networks and found that the patterns recognized by
+the algorithms could be re-used across tasks. Their aim was to analyze the free
+text portions of pathology reports to identify the primary site and laterality
+of tumors. The only features the authors supplied to the algorithms that they
+evaluated were unigrams and bigrams. These are the counts for single words and
+two-word combinations in a free text document. They subset the full set of words
+and word combinations to the 400 most commonly used ones. The machine learning
+algorithms that they employed (naive bayes, logistic regression, and deep neural
```
agitter: "Naive Bayes" or "naive Bayes"?

sections/03_categorize.md
```diff
@@ -28,7 +28,148 @@ we share these data while respecting privacy? I've made a section for each of
...
+feature construction methods [@doi:10.1038/srep26094]. While each of these
+touched on clinical tests, neither considered full text records. Taken together,
+these results support the potential of unsupervised feature construction in this
+domain. However, there are numerous challenges that will need to be overcome
+before we can fully assess the potential of deep learning for this application
+area.
+
+##### Opportunities
+
+However, significant work needs to be done to move these from conceptual
```
agitter: @cgreene are you planing to write these placeholder sections or are you looking for help here?

sections/03_categorize.md
```diff
@@ -28,7 +28,148 @@ we share these data while respecting privacy? I've made a section for each of
...
+advances to practical game-changers.
+
+##### Unique challenges
+
+Additionally, unique barriers exist in this space that may hinder progress in
+this field.
+
+###### Data sharing and privacy?
```
agitter: @XieConnect are you planning to write something about this topic?

sections/06_discussion.md
cgreene: I guess I prefer to introduce in specific sections and cross-domain recap here. I think it's going to be key in bio for many near-future advances.

sections/03_categorize.md
cgreene: We should settle on one. I don't have a preference either way. What do you prefer? I suggest we make one quick PR to go through and match case as soon as we decide.

sections/03_categorize.md
cgreene: Definitely. These are adversarial examples though. They take the same base image and apply perturbations to it. I think you're probably thinking of something more like this - which is an adversarial network generated training example:
https://arxiv.org/pdf/1612.07828v1.pdf

sections/03_categorize.md
cgreene: There are some features that have been developed in the past for this task. They first perform supervised learning where the task is to regress output nodes (if I recall correctly) towards those features. Then, after that process, they fine tune specifically on their supervised examples. This is perhaps a more efficient way (in terms of # of examples) to build the network by pushing it towards intermediate features that are thought to be useful.

sections/03_categorize.md
cgreene: I think the language in the paper was pretty loose. The pathologist had no false positives, so theirs are all false negatives. For the algorithm it had the normal FP/FN tradeoff. I think we may want to be less precise here. I'll make that revision now.

sections/03_categorize.md
cgreene: At this point, bullets are probably good. Still have some more work to do on this but wanted to give @sw1 a chance to integrate his new text.

sections/03_categorize.md
cgreene: Went with sentence case.

cgreene:
Based on @agitter I'm going to go ahead and merge this. This will allow @sw1 to have the token. @sw1  - please open a quick PR to add in your paragraph. Thanks!

sections/03_categorize.md
agitter: Got it.  You're right, my mind jumped to Generative Adversarial Networks when I saw adversarial.

PR #192, opened 2017-01-09
Update 03_categorize.md

sw1:
Survival analysis addition.


cgreene:
Hi @sw1 - can you reflow this to 80chars/line? This will allow comments at the line level. Atom can handle this automatically. Just select the text that you contributed and mark "reflow selection"

Thanks!

sections/03_categorize.md
```diff
@@ -144,7 +144,28 @@ domain. However, there are numerous challenges that will need to be overcome
...
+network that, in addition to outperforming Cox regression, was capable of 
+comparing the risk between a given pair of treatments, thus potentially acting 
+as recommender system (@arxiv:1606.00931). To overcome the remaining 
+difficulties, researchers have turned to deep exponential families, a class of 
+latent generative models that are constructed from any type of exponential 
+family distributions (@arxiv:1411.2581v1). The result was a deep survival 
+analysis model capable of overcoming challenges posed by missing data and 
+heterogeneous data types, while uncovering nonlinear relationships between 
+covariates and failure time. They showed their model more accurately 
+stratified patients as a function of disease risk score compared the current
```
cgreene: Are there specific challenges with using a deep neural network as opposed to current standard methods?

sections/03_categorize.md
```diff
@@ -144,7 +144,28 @@ domain. However, there are numerous challenges that will need to be overcome
...
+Still, recent work has also revealed domains in which deep networks have 
+proven superior to traditional methods. Survival analysis models the time 
+leading to an event of interest from a shared starting point, and in the 
+context of EHR data, often associates these events to subject covariates. 
+Exploring this relationship is difficult, however, given that EHR data types 
+are often heterogeneous, covariates are often missing, and conventional 
+approaches require the covariate-event relationship be linear and aligned to a 
+specific starting point (@arxiv:1608.02158). Early approaches, such as the
```
cgreene: Can you make these into square brackets for easy reference extraction?

cgreene: Also - just noticed the trailing spaces. While it's minor, can you trim those off at the same time you square bracket the references? Otherwise I will trim them later, but then it will be hard for readers to associate lines here with your work via `blame`. Thanks!

gwaygenomics:
@cgreene @sw1 are we good to merge this? @sw1 it looks like there is one additional question, but it  looks like square brackets are used

PR #201, opened 2017-01-13
Update 02_intro.md

AnneCarpenter:
Hi all, just joining in here and hoping to write some in a section on imaging. I thought it'd be a good idea to read the intro to get a sense of the paper first, and while I'm here I made edits! Mainly for readability and punchiness - aiming towards brevity.

PR #202, opened 2017-01-13
Wrote morphological phenotypes section in 04_study.md

AnneCarpenter:
As requested by @cgreene!

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -59,14 +59,45 @@ particularly notable in this area?*
 *There is not much content here.  Can [@tag:Chen2015_trans_species] be covered
 elsewhere?*
 
-### Cellular phenotypes
+### Morphological phenotypes
 
-*These are primarily imaging-based phenotypes.  We have not surveyed this area
-very comprehensively.  We could decide to not make imaging a primary focus,
-refer to existing reviews, and mention only a few particularly noteworthy
-representative papers.  Alternatively, we need to expand our literature review
-and summaries immediately if someone wants to be responsible for this
-sub-section.*
+A field poised for dramatic revolution by deep learning is 
+bioimage analysis - and not in the way you might first think. 
+Thus far, the primary use of deep learning for biological 
+images has been for segmentation - that is, for the identification 
+of biologically relevant structures in images such as nuclei, 
+infected cells, or vasculature. Once so-called regions of 
+interest have been identified, it is often straightforward 
+to measure biological properties of interest, such as fluorescence 
+intensities, textures, and sizes. Given the dramatic successes of 
+deep learning in biological imaging, we simply refer to recent reviews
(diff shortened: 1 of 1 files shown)


agitter:
Thanks for this and #201.  I'm backlogged in reviewing pull requests (teaching starts next week!) but will get to it if no one else does first.

AnneCarpenter:
Sure thing. BTW, given the stated goal for the article is to be predictive/interesting as opposed to exhaustively comprehensive,  I aimed for a conversational style. If it doesn't fit your vision feel free to modify.

brettbj:
@agitter - do we have a pull request process? I can read/review 1-2 but was leaving them to you and @cgreene (I know he just started teaching again too)

cgreene:
Please provide a review for anything that you are able to review. Anything
from content to grammar to flow with the other sections (probably not as
relevant now as it will be later). All of these items will make life easier
when the time comes to submit.

Thanks!

agitter:
I concur.  I think the spirit of the [peer review policy](https://github.com/greenelab/deep-review/blob/master/CONTRIBUTING.md#peer-review) (for those who haven't seen it) is that everyone can contribute reviews.  That will especially be helpful for topics where the repository maintainers lack expertise.

sections/04_study.md
```diff
@@ -59,18 +59,52 @@ particularly notable in this area?*
...
-and summaries immediately if someone wants to be responsible for this
-sub-section.*
-
-*Transfer learning from non-biological datasets to biological imaging
-data could fit here, and that does seem like an important topic.  Or
-transfer learning could be a more general topic for the Discussion section.*
+### Morphological phenotypes
+
+A field poised for dramatic revolution by deep learning is
+bioimage analysis - and not in the way you might first think.
```
cgreene: I'm a bit confused - what is the way that I might at first think? Is the point that:
* Deep learning + images is generally supervised.
* However sometimes we need to go fishing.
* Unsupervised deep learning approaches provide the opportunity to identify features that we don't know we should be looking for - this will be the revolution (and what is unexpected)?

If this is the idea, I'd suggest leading the reader through it a bit more. Maybe by first highlight how deep learning for supervised tasks has been great. Then come in later with - but, hey, this unsupervised stuff is the real game changer.

If this isn't the idea, then it went over my head. Maybe a bit more of a lead through the section would be helpful. I like the conversational tone. We may change it a bit later as we integrate the sections, but we'll probably also revise things that are less conversational to be more so.

AnneCarpenter: I see your point! Esp if you're trying to formalize the style, I'd suggest just deleting " - and not in the way you might first think" because indeed it is a little too tough to explain and it doesn't become clear until the next paragraph which is too long for someone to be confused. I'm going to try to figure out how to make this edit myself but if I fail, someone else should!

PR #212, opened 2017-01-26
Note repeatedly discussed topics for conclusions

cgreene:
Just starting to create a stub for this. I think these points have come up in >1 section (categorize, study, treat) now, so may belong in the conclusions. We can expand or pare this list as we move forward. I'd propose merging this (or editing + merging) to get some pieces in place.


cgreene:
@agitter : made a note that some of these might not make the cut. I think that a short punchy conclusion is the way to go. Thanks! :+1:

Issue #235, opened 2017-01-31
SVMs for TFBS Prediction


jacklanchantin:
http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1003711

**Abstract:** Any statistical learning approach using k-mers as features becomes susceptible to noisy training set k-mer frequencies once k becomes large. To address this problem, we introduce alternative feature sets using gapped k-mers, a new classifier, gkm-SVM, and a general method for robust estimation of k-mer frequencies. To make the method applicable to large-scale genome wide applications, we develop an efficient tree data structure for computing the kernel matrix. We show that compared to our original kmer-SVM and alternative approaches, our gkm-SVM predicts functional genomic regulatory elements and tissue specific enhancers with significantly improved accuracy, increasing the precision by up to a factor of two. We then show that gkm-SVM consistently outperforms kmer-SVM on human ENCODE ChIP-seq datasets, and further demonstrate the general utility of our method using a Naïve-Bayes classifier. Although developed for regulatory sequence analysis, these methods can be applied to any sequence classification problem.

http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1004271

**Abstract:** Genome-wide maps of transcription factor (TF) occupancy and regions of open chromatin implicitly contain DNA sequence signals for multiple factors. We present SeqGL, a novel de novo motif discovery algorithm to identify multiple TF sequence signals from ChIP-, DNase-, and ATAC-seq profiles. SeqGL trains a discriminative model using a k-mer feature representation together with group lasso regularization to extract a collection of sequence signals that distinguish peak sequences from flanking regions. Benchmarked on over 100 ChIP-seq experiments, SeqGL outperformed traditional motif discovery tools in discriminative accuracy. Furthermore, SeqGL can be naturally used with multitask learning to identify genomic and cell-type context determinants of TF binding. SeqGL successfully scales to the large multiplicity of sequence signals in DNase- or ATAC-seq maps. In particular, SeqGL was able to identify a number of ChIP-seq validated sequence signals that were not found by traditional motif discovery algorithms. Thus compared to widely used motif discovery algorithms, SeqGL demonstrates both greater discriminative accuracy and higher sensitivity for detecting the DNA sequence signals underlying regulatory element maps. 

**Reason I think these should be in the review: ** These were the first strong case for computational methods on TFBS prediction. Although they couldn't handle large amounts of sequences and needed well defined input features, they showed that machine learning could be accurately used for the task.

Issue #234, opened 2017-01-31
An assessment of neural network and statistical approaches for prediction of E. coli promoter sites.


jacklanchantin:
https://www.ncbi.nlm.nih.gov/pubmed/1508724

**Abstract:** We have constructed a perceptron type neural network for E. coli promoter prediction and improved its ability to generalize with a new technique for selecting the sequence features shown during training. We have also reconstructed five previous prediction methods and compared the effectiveness of those methods and our neural network. Surprisingly, the simple statistical method of Mulligan et al. performed the best amongst the previous methods. Our neural network was comparable to Mulligan's method when false positives were kept low and better than Mulligan's method when false negatives were kept low. We also showed the correlation between the prediction rates of neural networks achieved by previous researchers and the information content of their data sets.

**Reason I think it should be in the review:** This was the first paper to use a neural net for protein prediction. Their approach didn't work, but it shows why CNNs are better than a MLP for this task.

Issue #233, opened 2017-01-31
DNA binding sites: representation and discovery.


jacklanchantin:
**Abstract:** The purpose of this article is to provide a brief history of the development and application of computer algorithms for the analysis and prediction of DNA binding sites. This problem can be conveniently divided into two subproblems. The first is, given a collection of known binding sites, develop a representation of those sites that can be used to search new sequences and reliably predict where additional binding sites occur. The second is, given a set of sequences known to contain binding sites for a common factor, but not knowing where the sites are, discover the location of the sites in each sequence and a representation for the specificity of the protein.

**Reason I think it should be in the review:** This is a great summary of the task and computational approaches for predicting TFBSs. There's no sense in repeating all of what they said as for previous work.

agitter:
I like the idea of using a review to introduce what came before neural nets in #236.  This looks like it covers some essential topics like PWMs.  Do we need a second review to cover the some of the advances in the 2000 to ~2015 era(s)? 

I also edited the link above.  The link to Bioinformatics on PubMed is broken.

Issue #85, opened 2016-08-20
Deep Motif Dashboard: Visualizing and Understanding Genomic Sequences Using Deep Neural Networks


agitter:
http://arxiv.org/abs/1608.03644

> Deep neural network (DNN) models have recently obtained state-of-the-art prediction accuracy for the transcription factor binding (TFBS) site classification task. However, it remains unclear how these approaches identify meaningful DNA sequence signals and give insights as to why TFs bind to certain locations. In this paper, we propose a toolkit called the Deep Genomic Dashboard (Deep GDashboard) which provides a suite of visualization strategies to extract motifs, or sequence patterns from deep neural network models for TFBS classification. We demonstrate how to visualize and understand three important DNN models: convolutional, recurrent, and convolutional-recurrent networks. Our first visualization method is finding a test sequence's saliency map which uses first-order derivatives to describe the importance of each nucleotide in making the final prediction. Second, considering recurrent models make predictions in a temporal manner (from one end of a TFBS sequence to the other), we introduce temporal output values, indicating the prediction score of a model over time for a sequential input. Lastly, a class-specific visualization strategy finds the optimal input sequence for a given TFBS positive class via stochastic gradient optimization. Our experimental results indicate that a convolutional-recurrent architecture performs the best among the three architectures. The visualization techniques indicate that CNN-RNN makes predictions by modeling both motifs as well as dependencies among them.

@akundaje commented that using gradients in this manner will miss important patterns relative to #50.

akundaje:
They are using saliency maps (gradients) from Simonyan et al. with some
simple but interesting extensions for RNNs and CNN-RNNs. Nice paper. But
biggest drawback as we show in the Deeplift paper is that using gradients
can miss very clear predictive patterns in the sequences.

On Aug 20, 2016 4:46 AM, "Anthony Gitter" <notifications@github.com> wrote:

http://arxiv.org/abs/1608.03644

Deep neural network (DNN) models have recently obtained state-of-the-art
prediction accuracy for the transcription factor binding (TFBS) site
classification task. However, it remains unclear how these approaches
identify meaningful DNA sequence signals and give insights as to why TFs
bind to certain locations. In this paper, we propose a toolkit called the
Deep Genomic Dashboard (Deep GDashboard) which provides a suite of
visualization strategies to extract motifs, or sequence patterns from deep
neural network models for TFBS classification. We demonstrate how to
visualize and understand three important DNN models: convolutional,
[...]

PR #236, opened 2017-01-31
[WIP] 04_study: Transcription factors and RNA-binding proteins

jacklanchantin:
I'm writing the section on Transcription factors and RNA-binding proteins. I am planning on writing an intro paragraph explaining the task of predicting TFBSs and RNA-binding proteins, and then going into detail on each.

### Transcription Factor Proteins:
I think that below are the core papers related to TFBS prediction. I will start with a background explaining the task and ChIP-Seq. Then talk about the initial PWM consensus sequence/motif matching approaches, followed by SVMs, DeepBind, and the most deep learning approaches. I am planning on covering the DeepBind model, the flaws in its simple architecture, and then discuss deeper CNN models,  LSTMs, and CNN-LSTM models. 

Feel free to correct me, but in my opinion, a lot of the recent works make minor architectural changes, and claim a state-of-the-art model by showing a minor change in auROC score (which @akundaje pointed out isn't always the best). In many cases, it's really hard to tell what is "better" about the new model. I have seen this so commonly with computer vision and NLP papers, where someone may get a 0.1 increase in accuracy, but it's hard to tell if that's really a result of the model. That being said, I think it's hard to say what is "best" right now, but I do think that we captured a good summary of this in the Deep Motif Dashboard paper by explaining the key features of a CNN-LSTM, which works the best (which I can summarize here). I believe that the most important result is that pure motif matching isn't the best way to model TFBS. There are long range temporal dependencies.

In my experience with med school collaborations, the visualizations of the model predictions are just as important, if not more important than the individual predictions. #85 and #50 are two important visualization papers in this particular area. In terms of visualizing important nucleotides, #50 seems like a better approach than #85 . However, I think there are other visualization methods such as class optimization (i.e. motif generation) and temporal dependency visualizations, which I can discuss. I don't believe there is a "one size fits all" approach to visualizations. It depends on who is using the model.  @akundaje, what are your overall thoughts on this?

Next accomplishments: I think that so far, most of the work has been done on very defined ChIP-seq datasets where the sequences are all of the same length, they are doing binary prediction on a continuous ChIP-seq value, and they don't take into account multiple TFs binding in one place. In order to more accurately "simulate" functional genomic events, we need better prediction tasks and datasets. Considering a dataset that accounts for variable sequence length, does ChIP-seq regression rather than classification, and is a multi-class output, requires a much more complex model, but I think it needs to be considered.

- [ ] [DNA binding sites: representation and discovery (#233)](https://github.com/greenelab/deep-review/issues/233)
- [ ] [An assessment of neural network and statistical approaches for prediction of E. coli promoter sites. (#234)
](https://github.com/greenelab/deep-review/issues/234)
- [ ] [Enhanced Regulatory Sequence Prediction Using Gapped k-mer Features(#235)](https://github.com/greenelab/deep-review/issues/235)
- [ ] [SeqGL Identifies Context-Dependent Binding Signals in Genome-Wide Regulatory Element Maps
(#235)](https://github.com/greenelab/deep-review/issues/235)
- [ ] [Predicting the sequence specificities of DNA- and RNA-binding proteins by deep learning (#11)](https://github.com/greenelab/deep-review/issues/11)
- [ ] [Deep Motif Dashboard: Visualizing and Understanding Genomic Sequences Using Deep Neural Networks (#85)](https://github.com/greenelab/deep-review/issues/85)
- [ ] [Not Just a Black Box: Learning Important Features Through Propagating Activation Differences (#50)](https://github.com/greenelab/deep-review/issues/50)

### RNA-binding Proteins:

I am not as experienced with this as I have not done any experiments on this data, but I can try. Does anyone else have expertise in this. I know it's very similar to TFBS prediction.


gwaygenomics:
@jacklanchantin I think that outline sounds great! I was thinking about contributing to the study section as well, but moreso in the `gene expression` domain - which I think also ties nicely into gene regulation and TFBS/RBPs. So I thought I would jump right in on this PR!

A couple points: 
1. How have we been dealing with algorithmic solutions "pre-deep learning revolution" applied to the field? i.e. how much do we want to discuss PWMs and SVMs?
2. I think it also you may want to make sure your `treat-screening` branch is up to date with the most recent greenelab master branch as there will probably be less headache to get material merged. I think this is why all of these commits are appearing. The [Network Graph](https://github.com/jacklanchantin/deep-review/network) may help figuring this out.

jacklanchantin:
Thanks a lot for the comments @gwaygenomics. 

1. I think it's important to discuss PWMs in the sense that this is what biomedical researchers thought were the main influences of TFBSs. They believe that a TFBS is due to these consensus sequences which can be learned via PWMs in methods such as MEME. As for SVMs, I think that they are powerful in this domain if you have the right features, but the main issue is that they are so slow in testing. This is especially a problem when you may have to test hundreds of thousands of samples. 

2. Yes, you're right I completely screwed up the pull request. Let me fix that.

gwaygenomics:
sounds good @jacklanchantin - I agree that its important to give context into how computational predictions have evolved. I think it may also be equally as important to prognosticate how far deep learning can go - something which may be a theme or flavor of our review and I think was done quite elegantly in the [morphological phenotypes](https://github.com/greenelab/deep-review/blob/master/sections/04_study.md#morphological-phenotypes) subsection.

Once you get the branches synced and have pieces of your section written please let me know - I'd be happy to review. I can also let you know when I submit a PR of my subsection!

agitter:
@jacklanchantin your approach looks good to me.  You seem to be suggesting that there has been a short term "plateau" in performance, where architectural differences produce only small bumps in accuracy.  Am I reading that correctly?

We'll have a separate section on problems with ROC for evaluation with imbalanced classes so feel free to refer to that (to-be-written) section here.

For this review in particular, I think we should mention the line of work that seeks to use TF binding prediction to interpret non-coding genetic variation.  That could be a separate pull request though because you already have many papers to cover here.

cgreene:
I like the path that you are taking with this. I think there is something a bit odd going on with this PR since it looks like any commits to master are getting moved to treat-screening. @jacklanchantin - Is this desired or should we re-setup this PR?

If you make your own fork, and then try to PR from your own fork into `greenelab/deep-review master` that might be the way to go.

jacklanchantin:
Yeah, sorry I screwed up this PR. I need to reset it

jacklanchantin:
@agitter . Yes, I agree. That would be along the lines of the DeepSEA paper right?

agitter:
@jacklanchantin Yes, DeepSEA is one example.  There are others as well, though I don't recall which use deep learning and which use other approaches.

Either way, I think we can leave genetic variant interpretation as a TODO and work on merging a draft of this first.

PR #239, opened 2017-02-02
Fix Reference Tags

gwaygenomics:
Adding tabs where they were previously missing


agitter:
Looks good, I'll merge it.

PR #245, opened 2017-02-17
Fix references and add basic reference extracting script

dhimmel:


agitter:
Looks good.  I'm merging this.

PR #253, opened 2017-02-24
Fix broken references

dhimmel:
Was a real pain to fix all these references. Will aim to get CI set up soon, so builds will fail if the citations are faulty.


agitter:
Thanks for fixing all of these.

PR #259, opened 2017-02-25
Continuously analyze repo for automatic manuscript deployment

### LINKED ISSUE #255: How should we set up Continuous Integration / Analysis
Soon I will complete python infrastructure to automatically download bibliographic metadata for each citation. Pandoc will be able to read in this data to automatically generate the bibliography.

I was planning on using Travis CI to automatically run the bibliographic conversions on each pull request and new commit. This should prevents the type of errors that were fixed by #253. In other words, the build will fail when poorly formatted reference is added. We can also make builds fail if a citation lacks the expected metadata.

However, Travis can also run the Pandoc conversion to create a HTML and PDF outputs. I'd love to use [continuous analysis](https://doi.org/10.1101/056473) to also commit these output files (and possibly the intermediate files), back to the repo. Just not sure what the best setup is:

+ Do we push back to the tip of a pull request?
+ Do we only commit back once a pull request is merged?
+ What should our branching model be?

We should probably also host the final HTML page via GitHub pages, which corresponds with the `gh-pages` branch.

Paging the experts: @brettbj @cgreene @agitter

dhimmel:
Closes #255.


dhimmel:
Not sure of how to test this without merging it... so here we go!

Issue #269, opened 2017-03-03
Use OpenTimestamps to prove manuscript existence at time of build


dhimmel:
http://opentimestamps.org is a free and easy (but still nascent) method for proving file existence at a certain time. OpenTimestamps allows you to encode the hash of a file in the Bitcoin blockchain, the must immutable database known to humankind.

The way I was thinking of implementing OpenTimestamps is to have our [deployment script](https://github.com/greenelab/deep-review/blob/6f6c1fc83240984fe9414743ace998c37b23e5b9/ci/deploy.sh) stamp `index.html` and `deep-review.pdf` when they get automatically generated. We would then also save the `.ots` which are needed for verification.

PR #274, opened 2017-03-03
Timestamp manuscripts with OpenTimestamps

dhimmel:
Uses opentimestamps.org to prove existence of each manuscript version at their corresponding points in time.

Runs for deployment manuscripts only.

Closes https://github.com/greenelab/deep-review/issues/269

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Generated manuscript output files
 output/index.html
 output/deep-review.pdf
+output/*.ots
 
 # Generated reference files
 references/generated/*
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "ci/opentimestamps-client"]
+	path = ci/opentimestamps-client
+	url = https://github.com/opentimestamps/opentimestamps-client.git
--- a/build/environment.yml
+++ b/build/environment.yml
@@ -12,3 +12,4 @@ dependencies:
     - arxiv2bib==1.0.7
     - bibtexparser==0.6.2
     - ghp-import==0.5.5
+    - python-bitcoinlib==0.7.0
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -1,3 +1,9 @@
+# Generate OpenTimestamps
+python ci/opentimestamps-client/ots stamp \
+  output/index.html \
+  output/deep-review.pdf
+
+# Exit on errors
 set -o errexit
 
 # Configure git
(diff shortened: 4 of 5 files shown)


dhimmel:
`greenelab/deep-review` is a collaboratively written review article on deep learning in precision medicine. We're using GitHub to write the article in markdown. We have a build process to automatically download citation medata, create the bibliography, and export HTML and PDF versions of the manuscript. We use Travis CI to:

1. Build every pull request, thereby spotting citation or other errors.
2. Build commits to `master` and deploy (commit) the resulting manuscript to and citation information back to the GitHub.

This pull request adds manuscript proof of existence to part 2 above. It uses [OpenTimestamps](https://github.com/opentimestamps/opentimestamps-client) to create manuscript timestamps (manuscirpt hashes that go into a Merkle tree whose root is written to the Bitcoin blockchain). The `.ots` files that are created, allow verification at later dates.

OpenTimestamps by @petertodd is very innovative... it's low cost to run and scales since users store the relevant path through the Merkle tree themselves. OpenTimestamps is still alpha, so it may not be all smooth sailing.

Alerting @ctb and @profjsb, since we discussed proof of existence for science at the Moore Investigator Symposium. Also tagging @bgcarlisle who's [blogged about](http://www.bgcarlisle.com/blog/2014/08/25/proof-of-prespecified-endpoints-in-medical-research-with-the-bitcoin-blockchain/) timestamping science using bitcoin. Thoughts and code review appreciated.

dhimmel:
Some potential shortcomings of the current implementation:

+ only manuscript existence is proven, not who contributed what parts. Who contributed what at what time would be more useful. If we timestamp the git commit hash of the source repository, then we could also cryptographically timestamp the commit history. Note that git's reliance on the [recently broken](https://arstechnica.com/security/2017/02/at-deaths-door-for-years-widely-used-sha1-function-is-now-dead/) SHA1 is not ideal.
+ it doesn't prove that no other versions of the manuscript existed at the same time. While this won't matter for this review, one could imagine a researcher writes many documents, timestamps them, and then later only publicizes the documents that turned out to be true.

dhimmel:
> If we timestamp the git commit hash of the source repository, then we could also cryptographically timestamp the commit history.

https://github.com/greenelab/deep-review/pull/274/commits/f39ba3baf0da184608863feee69c7509e402943f adds the git commit hash to a README file, which is stamped.

dhimmel:
Going to merge this now as it's hard to test the Travis deployment without actually pushing a change to `master`.

But discussion still welcome.

Issue #210, opened 2017-01-25
A predictive model of gene expression using a deep learning framework


agitter:
https://doi.org/10.1109/BIBM.2016.7822599

> With an unprecedented amount of data available, it is important to explore new methods for developing predictive models to mine this data for scientific discoveries. In this study, we propose a deep learning regression model based on MultiLayer Perceptron and Stacked Denoising Auto-encoder (MLP-SAE) to predict gene expression from genotypes of genetic variation. Specifically, we use a stacked denoising auto-encoder to train our regression model in order to extract useful features, and utilize the multilayer perceptron for backpropagation. We further improve our model by adding a dropout technique to prevent overfitting. Our results on a real genomic dataset show that our MLP-SAE model with dropout outperform Lasso, Random Forests, and MLP-SAE without dropout. Our study provides a new application of deep learning in mining genomics data, and demonstrates that deep learning has great potentials in building predictive models to help understand biological systems.

gwaygenomics:
I couldn't find the full text for this paper, anyone else have any luck?

agitter:
I get a "View Document" link directly under the title when I visit http://ieeexplore.ieee.org/document/7822599/.

A quick skim shows it to be a fairly standard model.  Input layer (features: SNP genotypes) -> autoencoder layer 1 -> autoencoder layer 2 -> output layer (regression).  They implement the model in pylearn2, but I don't see code.

They model yeast data from https://doi.org/10.1073/pnas.0408709102, which has 112 microarray samples.  "We extract the genotypes of 2,956 SNPs and gene expression profiles of 7,085 genes in the 112 samples", where I'm going to infer "genes" might mean "probes".   Performance is evaluated with mean squared error (MSE) in 5-fold cross validation.  They compare to L1-regularized linear regression and random forests.  I don't see strong evidence the neural network beats linear regression in any meaningful way.  The MSEs are almost the same under the best hyperparameters: 0.291 (LR) vs. 0.289 (NN).  Not to mention ~0.3 seems like a high MSE.

I'm closing this and will suggest we don't include it unless someone wants to convince me otherwise.

Issue #153, opened 2016-12-08
Removal of Batch Effects using Distribution-Matching Residual Networks


sw1:
https://arxiv.org/abs/1610.04181

> Sources of variability in experimentally derived data include measurement error in addition to the physical phenomena of interest. This measurement error is a combination of systematic components, originating from the measuring instrument, and random measurement errors. Several novel biological technologies, such as mass cytometry and single-cell RNA-seq, are plagued with systematic errors that may severely affect statistical analysis if the data is not properly calibrated. Here, we propose a novel deep learning approach for removing systematic batch effects. Our method is based on a residual network, trained to minimize the Maximum Mean Discrepancy (MMD) between the multivariate distributions of two replicates, measured in different batches. We apply our method to mass cytometry and single-cell RNA-seq datasets, and demonstrate that it effectively attenuates batch effects, and outperforms several popular methods.

They used a deep residual net with MMD loss to normalize data with batch effects. They assumed the distributions are somewhat close to each other.

This might be a little off topic for the overall scope of the review, but it could certainly does have medical applications.

Issue #24, opened 2016-08-05
Gene expression inference with deep learning


cgreene:
https://doi.org/10.1093/bioinformatics/btw074

gwaygenomics:
Great example of deep learning (feed forward neural network) significantly outperforming a simpler machine learning algorithm (linear regression) on an important task (predicting gene expression from [an informative panel](http://www.lincscloud.org/l1000/ "LINCS 1000") Also demonstrates ability of classifier trained on microarray data to infer RNAseq data.

### Biology
* Inferring expression of ~10,000 genes using only the measurements of a representative set of ~1000 genes
* Comparing performance of algorithm to infer RNAseq vs. Microarray assayed genes
* Trained on GEO data and evaluated on holdout GEO, GTEx, and 1000 genomes expression data

### Computational aspects
* Feed forward neural network
  * Trained with dropout, momentum, adaptive learning rate and with glorot initialized weights
  * Three hidden layers
    * 9000 hidden units in each layer shown to perform best
* Performance evaluated using mean absolute error (MAE) at a _per gene_ level
  * Compared to standard LINCS method (linear regression) and KNN regression
* Nice discussion of visualizing network components

### Results
* ~15% improvement in MAE in NN compared to LR, KNN regression performed poorly
* 99.97% of genes show improvement in NN method compared to LR
* Trained NN on microarray data predicts RNAseq genes very well
  * Different optimal hyperparameters
    * Two hidden layers (9000 x 2)
  * NN does not have as significant improvements (but still improvements!) compared to LR
* Visualization techniques
  * Sparse connections from input layer to hidden layer
  * Certain hidden layers are _hubs_ of activity - probably capture some central co-regulated organization principles

agitter:
> Ideally, we should also configure D-GEX with 9520 units in the output layer corresponding to the 9520 target genes. However, each of our GPUs has only 6 GB of memory, thus we cannot configure hidden layers with sufficient number of hidden units if all the target genes are included in one output layer. Therefore, we randomly partitioned the 9520 target genes into two sets that each contains 4760 target genes. We then built two separate neural networks with each output layer corresponding to one half of the target genes.

This part was unfortunate.  I wonder how much better they could have done without this artificial limitation.

Their [code](https://github.com/uci-cbcl/D-GEX) has pairs of scripts for training networks on the first half of the data and then the second half.  It may not be too difficult to train on all of the genes if someone is feeling especially curious (not me).

gwaygenomics:
[...]

Issue #19, opened 2016-08-05
Learning a hierarchical representation of the yeast transcriptomic machinery using an autoencoder model 


cgreene:
https://dx.doi.org/10.1186/s12859-015-0852-1

agitter:
The precise biological meaning ascribed to each of the hidden layers in Figure 1c may risk overinterpreting the model.

cgreene:
@agitter : Figure 3 is surprisingly clean for the first layer. I agree that the evidence for the later levels in 
"The hierarchical structure captures signals of different degrees of abstraction" seems a bit much for the amount of evidence. Probably good to discuss alongside #6 #22 & the rest in a nucleus of gene expression + unsupervised feature construction papers.

cgreene:
Include with unsupervised feature construction. May also be a useful example for of potential dangers of too much interpretation cc @akundaje

gwaygenomics:
Just revisited this paper - I'm going to reopen it to cite in the gene expression study section (It doesn't appear that we've stuck to the "closing an issue when its not being discussed" paradigm). I agree with the over interpretation concerns, but the concept is very cool. The paper also had a nice discussion of limitations of current approaches.

# Biological Aspects

* Yeast cDNA microarrays (1609 from puma.princeton.edu and 300 from [Hughes et al 2000](http://dx.doi.org/10.1016/S0092-8674(00)00015-5))
* Found 2228 genes to be differentially expressed between control and treatments (variety of treatments; chemical, gene mutation, etc.)
* Binary entries 0 or 1 if gene was differentially expressed or not

# Computational Aspects

* Compared NMF vs. Autoencoder vs. Sparse Autoencoder
* 4 hidden layers for autoencoders
* Evaluation based on reconstruction error and BIC
* Also evaluated how well Autoencoder learned features performed in a clustering analysis compared to raw and NMF features (Fig 5)

PR #286, opened 2017-03-30
Gene exp study

gwaygenomics:
Adding draft for gene expression component of study section. Comments welcome!


gwaygenomics:
also tagging @jacklanchantin for review

references/tags.tsv
```diff
@@ -48,11 +51,15 @@ Schatz2010_dna_cloud	doi:10.1038/nbt0710-691
...
 Seide2014_parallel	doi:10.1109/ICASSP.2014.6853593
 Shaham2016_batch_effects	arxiv:1610.04181
+Singh2016_deepchrome	arxiv:1607.02078
 Spark	doi:10.1145/2934664
 Stein2010_cloud	doi:10.1186/gb-2010-11-5-207
 Su2015_gpu	arxiv:1507.01239
 Sun2016_ensemble	arxiv:1606.00575
 Swamidass2009_irv	doi:10.1021/ci8004379
+Tan2014_psb	doi:10.1142/9789814644730_0014
+Tan2015_adage	doi:0.1128/mSystems.00025-15
```
agitter: DOI typo, it should be `10.1128/mSystems.00025-15`

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+approaches are promising, many convert gene expression measurements to
+categorical or binary variables thus ablating many complex gene expression
+signatures present in intermediate and relative numbers.
+
+Deep learning applied to gene expression data is in its infancy but the future
+is bright. Many hypotheses can now be interrogated because of increasing
+amounts of data and new developing technologies. For exampe, there is a growing
+appreciation for the large impact of disease heterogeneity on research and
+treatment strategies for disease. New technologies are being developed, such as
+single cell RNAseq and high throughput flourescence based imaging. These
```
agitter: "fluorescence", and perhaps "RNA-seq"

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+technologies are growing in popularity and will provide increasingly
+novel perspectives with respect to how cellular and spatial heterogeneity
+impacts gene expression coordination within a sample. Deep learning is already
+well established in the image processing community, so the marriage of
+flourescence based imaging techniques and deep learning is natural. In general,
```
agitter: "fluorescence"

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
-*Predicting gene expression levels and unsupervised approaches for learning
-from gene expression.  Those could be divided into separate sub-sections.*
+Gene expression measurements characterize the abundance of many thousands of
+RNA transcripts within a given organism, tissue, or cell. This characterization
+can represent the underlying state of the given system and can be used to study
+heterogeneity across samples as well as how the system reacts to perturbation.
+While gene expression measurements have been traditionally made by quantitative
+PCR, low throughput flourescense based methods, and microarray technologies, the
+field has shifted in recent years to primarily performing RNA sequencing to
+catalog total RNA. As such next generation sequencing technologies continue to
```
agitter: Remove "total" here?  It makes me think of the specific [total RNA-seq protocol](https://www.illumina.com/techniques/sequencing/rna-sequencing/total-rna-seq.html), I don't think that is the intention.

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+fall in price and rise in throughput, studying gene expression through the lens
+of deep learning will increase in practicality and promise.
```
agitter: Does this imply that the limiting factor now is the number of samples available?

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+of deep learning will increase in practicality and promise. 
+
+Already several deep learning approaches have been applied to gene expression
+data with varying aims. For instance, many researchers have applied unsupervised
+deep learning approaches to extract meaningful representations of gene modules
+or sample clusters. Unsupervised approaches, such as denoising autoencoders,
+have been applied to cluster yeast expression microarrays into known modules
+representing cell cycle processes [@tag:Gupta2015_exprs_yeast] and also to
+stratify yeast strains based on chemical and mutational perturbations
+[@tag:Chen2016_exprs_yeast]. Shallow (one hidden layer) denoising autoencoders
```
agitter: We had raised overinterpretation concerns in #19.  Should we introduce those here?

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+have also been fruitful in extracting biological insight from thousandas of
+_Pseudomonas aeruginosa_ experiments [@tag:Tan2015_adage @tag:Tan2016_eadage]
+and in aggregating features relevant to specific breast cancer subtypes
+[@tag:Tan2014_psb]. Alternatively, deep learning approaches are being applied
```
agitter: I'd like to see another line describing why these methods have been fruitful, e.g., what was learned.  This may be a good place to mention advantages of the unsupervised approach.  If you add a line or two here, "Alternatively" could start a new paragraph.

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+for gene expression prediction tasks. For example, a deep neural network with
+three hidden layers outperformed linear regression in inferring the expression
+of over 20,000 target genes based on a representative, well-connected set of
+about 1,000 landmark genes [@tag:Chen2016_gene_expr]. Furthermore, a
```
agitter: We had some critical questions about whether beating linear regression in #24 really makes a difference.  @dhimmel wrote

> While I think this paper makes the case that deep learning is better at imputation, I don't think it's good enough to salvage the imputed LINCS L1000 expression calls.

and that is an important point.  It is similar to the concern I had in #210, but there the neural network didn't even beat the baselines by much.

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+convolutional neural network applied to histone modifications
+[@tag:Singh2016_deepchrome] was shown to predict raw gene expression output.
```
agitter: I didn't read Singh2016.  Did it work well?  It would be valuable if we can produce an overall assessment about the state of the art in expression prediction and how far we are from that being a solved problem (though there have already been 10+ years of papers saying it is "solved").

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+Deep learning approaches have also been applied to study cancer gene expression
+data with goals of identifying subtypes of patients with different molecular
+features and clinical manifestations [@tag:Liang2015_exprs_cancer]. While these
```
agitter: Presumably this is supervised, in contrast with the unsupervised method above?  Is there anything worthwhile to compare between the supervised / unsupervised.

sections/04_study.md
agitter: I didn't read #153 about batch effects, but does that fit in here?

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+approaches are promising, many convert gene expression measurements to
+categorical or binary variables thus ablating many complex gene expression
+signatures present in intermediate and relative numbers.
+
+Deep learning applied to gene expression data is in its infancy but the future
```
agitter: Does that mean that if you were forced to make a "transformative" assessment you would say it isn't yet?

sections/04_study.md
```diff
@@ -27,8 +27,55 @@ dive into merits or weaknesses of individual methods.*
...
+appreciation for the large impact of disease heterogeneity on research and
+treatment strategies for disease. New technologies are being developed, such as
+single cell RNAseq and high throughput flourescence based imaging. These
+technologies are growing in popularity and will provide increasingly
+novel perspectives with respect to how cellular and spatial heterogeneity
+impacts gene expression coordination within a sample. Deep learning is already
+well established in the image processing community, so the marriage of
+flourescence based imaging techniques and deep learning is natural. In general,
+as the flow of gene expression data increases, the quality and types of
+questions deep learning can address is poised to improve.
```
agitter: This gets at what would be needed to improve gene expression prediction and related tasks, which could be a nice speculative question for us to address.  I'm not sure I agree that sample sizes are the only, or even the most prominent, limiting factor.  My wild speculation would propose that gene expression prediction models might need to integrate more information about the state of cells and their environments, perhaps in the spirit of [whole cell computational models](http://doi.org/10.1016/j.cell.2012.05.044).  You don't have to agree, but I'm curious what you think.

gwaygenomics:
Thanks @agitter and @jacklanchantin for very insightful comments. I addressed some of the minor suggestions (like spelling errors/typos). I will take some more time to think about some of the other comments - new commit and specific responses coming soon

sections/04_study.md
gwaygenomics: I think we always want more data - I didn't mean to imply that it is currently a limiting factor, although in some cases it definitely is. (E.g. studying gene expression of rare cancer types)

Instead, I meant to convey that increasing amounts of data make deep learning approaches more feasible (practicality) and that the pace of deep learning approaches applied to gene expression data is likely to grow leading to new discoveries and new methods (promise)

If you agree with that message, I will work on getting the point across more clearly.

Perhaps something like:

```
As such next generation sequencing technologies continue to fall in price and rise in throughput,
applying deep learning to study gene expression data is likely to make training more feasible. With increased modeling ability, deep learning approaches will grow in popularity and lead to novel biological insights.
```

sections/04_study.md
gwaygenomics: yes, i definitely think we should. I will update in the next commit

sections/04_study.md
gwaygenomics: I revisited this paper again and agree with this point. I will raise the distinction in a future commit!

sections/04_study.md
gwaygenomics: Based on Figure 3, it looks like the method _did_ consistently outperform other methods and predictions appear to be quite good (although also highly variable across different cell types). 

I am not too familiar with this field outside of this one paper, maybe @jacklanchantin will have some more insight here!

sections/04_study.md
gwaygenomics: Yeah, i think this would be an important point to mention

sections/04_study.md
gwaygenomics: Yeah, I would say its not there yet - but these early studies have overcome several limitations of alternative approaches and do improve performance

sections/04_study.md
gwaygenomics: great point! I just read this paper and will add my comments to the issue and pull request soon.

sections/04_study.md
gwaygenomics: > needed to improve gene expression prediction and related tasks

agree about data integration would improve gene expression prediction. However, when the data doesn't exist, its nice to have alternative methods that don't require many measured features.

I think in some ways the whole cell computational model is the polar opposite of deep learning approaches. While both are aiming to describe and predict cellular state, the way each arrives at the conclusion is different. The whole cell model tries to model everything in a closely supervised fashion incorporating currently known biology, while deep learning lets the data explain connections. I think figuring out better ways of interpreting deeply learned features would be necessary for it to move closer to whole cell models - but in a lot of cases, depending on the task, feature interpretation doesn't matter too much

gwaygenomics:
Thanks again @agitter and @jacklanchantin - I updated the draft addressing previous comments. The pull request is ready again for your review

agitter:
Thanks for making changes.  This looks good to me so I'll squash and merge.

Issue #238, opened 2017-02-01
Integrative Deep Models for Alternative Splicing


agitter:
https://doi.org/10.1101/104869

> Advancements in sequencing technologies have highlighted the role of alternative splicing (AS) in increasing transcriptome complexity. This role of AS, combined with the relation of aberrant splicing to malignant states, motivated two streams of research, experimental and computational. The first involves a myriad of techniques such as RNA-Seq and CLIP-Seq to identify splicing regulators and their putative targets. The second involves probabilistic models, also known as splicing codes, which infer regulatory mechanisms and predict splicing outcome directly from genomic sequence. To date, these models have utilized only expression data. In this work we address two related challenges: Can we improve on previous models for AS outcome prediction and can we integrate additional sources of data to improve predictions for AS regulatory factors. We perform a detailed comparison of two previous modeling approaches, Bayesian and Deep Neural networks, dissecting the confounding effects of datasets and target functions. We then develop a new target function for AS prediction and show that it significantly improves model accuracy. Next, we develop a modeling framework to incorporate CLIP-Seq, knockdown and over-expression experiments, which are inherently noisy and suffer from missing values. Using several datasets involving key splice factors in mouse brain, muscle and heart we demonstrate both the prediction improvements and biological insights offered by our new models. Overall, the framework we propose offers a scalable integrative solution to improve splicing code modeling as vast amounts of relevant genomic data become available. Availability: code and data will be available on Github following publication.

gwaygenomics:
Clear and comprehensive analysis making several comparisons against previous models. The authors nicely demonstrate how a deep neural network integrating diverse data types (such as [CLIP-seq](https://en.wikipedia.org/wiki/CLIP) and gene knockout) with genomic sequence features to predict alternative splice sites improves state of the art prediction. They compare this model to Bayesian neural networks and a previous model discussed in #4. Additionally, because of increasing amounts of available data and algorithm capacity, the authors are also able to develop a target function to model exon inclusion directly instead of categorizing exon inclusion as low, mid, or high.

# Biological Aspects

* Predict alternative splicing events from sequence features
  * Also determine if integrating CLIP-seq and knockout/knockdown data improve model
* Compare sequence feature based models to "ground truth" RNAseq inferred exon inclusion

# Computational Aspects

* Thorough comparison of four models
  * Deep Bayesian Network
  * Deep Neural Network implemented in #4 
  * Deep Neural Network with CLIP-seq integration
[...]

Issue #258, opened 2017-02-25
Imputation for transcription factor binding predictions based on deep learning


agitter:
http://doi.org/10.1371/journal.pcbi.1005403

> Understanding the cell-specific binding patterns of transcription factors (TFs) is fundamental to studying gene regulatory networks in biological systems, for which ChIP-seq not only provides valuable data but is also considered as the gold standard. Despite tremendous efforts from the scientific community to conduct TF ChIP-seq experiments, the available data represent only a limited percentage of ChIP-seq experiments, considering all possible combinations of TFs and cell lines. In this study, we demonstrate a method for accurately predicting cell-specific TF binding for TF-cell line combinations based on only a small fraction (4%) of the combinations using available ChIP-seq data. The proposed model, termed TFImpute, is based on a deep neural network with a multi-task learning setting to borrow information across transcription factors and cell lines. Compared with existing methods, TFImpute achieves comparable accuracy on TF-cell line combinations with ChIP-seq data; moreover, TFImpute achieves better accuracy on TF-cell line combinations without ChIP-seq data. This approach can predict cell line specific enhancer activities in K562 and HepG2 cell lines, as measured by massively parallel reporter assays, and predicts the impact of SNPs on TF binding.

@jacklanchantin can you please look at this for #236?

One minor comment is that they say
> Because the released DeepBind software package does not contain the training step, we could not train it on our dataset.

but I believe the DeepBind (#11) code [here](http://tools.genes.toronto.edu/deepbind/nbtcode/) does include the training step.  It's just hard to find from the [main page](http://tools.genes.toronto.edu/deepbind/).

jacklanchantin:
@agitter thanks for posting this. I actually don't think that the code they provide for deepbind includes the model construction. They include pre-trained models, which you can use to score new sequences. But the deepbind model is very straightforward, I don't know why they didn't just code it themselves.

agitter:
It's not important enough for me to download and check their code, but I saw text in the DeepBind [README ](http://tools.genes.toronto.edu/deepbind/nbtcode/README.TXT) like

> Train TF models on ENCODE ChIP-seq peaks, then test on held-out subset of peaks.
See supplementary information if training/testing set is not clear from descriptions below.
Use top 500 even to train, top 500 odd to test:
`python deepbind_train_encode.py top calib,train,test,report`

I believe there are two versions of the code, only one of which supports training.

knowledgefold:
[...]

PR #288, opened 2017-04-04
Initial draft of splicing

bdo311:
Would welcome comments and suggestions! I think it's also a little long. 

Used papers #238, #7

I can't seem to request a reviewer from the right sidebar... @agitter @cgreene

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -32,7 +32,81 @@ from gene expression.  Those could be divided into separate sub-sections.*
 
 ### Splicing
 
-*A separate section from general gene expression section above.*
+Pre-mRNA transcripts can be spliced into different isoforms by retaining or
+skipping subsets of exons, or including parts of introns. This alternative
+splicing provides cells with enormous spatiotemporal flexibility to generate
+multiple distinct proteins from a single gene. Splicing is catalyzed by small
+nuclear RNAs (snRNAs) and spliceosomal proteins, which detect sequence motifs
+such as splice sites and exon sequence enhancers and silencers (ESE and ESS).
+Various RNA-binding proteins and noncoding RNAs can bias these reactions by
+altering binding affinities, blocking splice sites, or sequestering splicing
+factors. This remarkable complexity lends itself to defects that underlie many
+diseases [@tag:Scotti2016_missplicing]. For instance, in Becker muscular
+dystrophy, a point mutation in dystrophin creates an ESS that induces skipping
+of exon 31. A recent study found that quantitative trait loci (QTLs) that affect
+splicing in lymphoblastoid cell lines are enriched within risk loci for
+schizophrenia, multiple sclerosis and other immune diseases, implicating
+mis-splicing as a much more widespread feature of human pathologies than
+previously thought [@tag:Li2016_variation].
+
+Sequencing studies routinely return thousands of unannotated variants. Which
+cause functional changes in splicing, and if so, how? Prediction of a “splicing
+code” has been a holy grail over the past decade. Initial machine learning
+approaches used a naive Bayes model and a 2-layer Bayesian neural network with
+thousands of hand-derived sequence-based features to predict the probability of
+exon skipping [@tag:Barash2010_splicing_code @tag:Xiong2011_bayesian]. With the
+advent of deep learning, more complex models were built that provided
+significantly better performance [@tag:Xiong2015_splicing_code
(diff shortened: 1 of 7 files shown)


agitter:
Thanks!  @cgreene @gwaygenomics or I will check it out and provide comments.  I'm not worried about the length.

I did see that the Travis CI build failed reported a problem with `@tag:Xiong2011_bayesian`.  There may be some leftover text in the `tags.tsv` files from a merge as well.

agitter:
@bdo311 I fixed the problems with the references and will have time to give more feedback today or tomorrow.

bdo311:
Thanks @agitter! I was fruitlessly trying to figure out what was wrong with
the Xiong citation...

sections/04_study.md
```diff
@@ -105,7 +105,81 @@ learning can address is poised to improve.
...
+The massive improvement seen with deep learning seems to stem from hidden layers
+being able to create new higher-order “features”, whereas earlier approaches
+often assumed independence of features and were unable to generalize.
+Higher-order understanding is especially important in splicing, which depends
+not only on the primary sequence, but also local RNA structure, tissue identity,
+splicing factor binding, and other currently unknown factors — all of which
+interact in complex, incompletely characterized ways. With new tools to
+interpret these meta-features, a major focus of current deep learning research,
+we will soon have the ability to extract a more nuanced biological understanding
+of splicing — perhaps by building different deep neural networks for different
```
agitter: I'm curious about the idea of building different networks for different conditions.  Is there something condition-specific that could not be learned from other types of input data?  Does the cell line embedding idea used in #258 have any relevance here?

I'm not disagreeing with the sentiment, but I found it thought-provoking and was hoping you could elaborate.

sections/04_study.md
```diff
@@ -105,7 +105,81 @@ learning can address is poised to improve.
...
+
+A parallel effort has been to use more data with simpler models. An exhaustive
+study using readouts of splicing for millions of synthetic intronic sequences
+was able to describe motifs that influence the strength of alternative splice
+sites [@tag:Rosenberg2015_synthetic_seqs]. Interestingly, a simple linear model
+built from this data successfully generalized to exon skipping: in a limited
+analysis using SNPs from three genes, it predicted exon skipping with three
+times the accuracy of Xiong et al.’s framework. This case is instructive in that
+clever sources of data, not just more powerful models, can lead to novel
+insights.
```
agitter: What type of model is used in Rosenberg2015?  Is it a neural network or something else?  Or is the the simple linear model described in the next sentence?

I'm wondering whether the main conclusion is that
1) the algorithms (neural networks) haven't been the bottleneck in exon skipping prediction but rather the data
2) better data provides future opportunities to train even better algorithms (e.g. neural networks)
3) I missed the point entirely.

bdo311: I feel like it's both. Different/complementary kinds of data provide opportunities to build more informative models, and better algorithms let you do more with integrating more diverse sources of data. I don't know if anyone has a definite answer on this or whether there should be a definite answer..

bdo311: The Rosenberg model is a simple linear model.

sections/04_study.md
bdo311: I was thinking that the hidden nodes learned by a network trained on brain data might tell you something about brain when compared to, for example, hidden nodes learned by a network trained on heart data. 

Yeah #4 and #258 use a similar approach where they encode cell type in the input. One of the benefits to this multi-task learning is that learning is more efficient and that you can exploit similarities (rather than having to learn them individually) -- there are some papers (https://arxiv.org/pdf/1701.08734.pdf) that take this idea to its extreme. Downside is that maybe building a jack-of-all-trades model will reduce its single-task prediction accuracy. Maybe I'll drop in a sentence about this.

bdo311: #4 and #7 (at least to my reading) use similar if not identical deep learning networks; they were developed by the same group (Frey's lab at UofT). So I chose just to cite the 2015 Science paper instead.

agitter:
Thanks for the responses to my comments.  I'm merging this.  When we make a second pass for further editing we can see if we want to say anything about the new paper #293.

@bdo311 some of your commits were authored with an email address that is not associated with your GitHub account.  I think it should be okay, and I'll fix authorship after I squash and merge if needed.

agitter:
Confirming that the squashed commit was correctly attributed to @bdo311

PR #297, opened 2017-04-07
Integrate recent contributions to categorize

cgreene:
The goal of this PR is to integrate contributions from @XieConnect and @traversc into the flow of the categorize section. Both @XieConnect and @traversc : I'd love to hear if you think this is ok. @traversc - I have a quick `TODO` for you to address after this merges. I hope it is relatively quick.

### DIFF
--- a/sections/03_categorize.md
+++ b/sections/03_categorize.md
@@ -228,26 +228,33 @@ provide one path forward [@doi:10.1016/j.jbi.2016.10.007], as do adversarial
 training examples [@doi:10.1101/095786]. We also expect the recently described
 data programming strategy [@arxiv:1605.07723] to play a role in addressing this
 challenge. In data programming, noisy automated labeling functions are
-integrated. While numerous writers have described data as the new oil [@url:http://ana.blogs.com/maestros/2006/11/data_is_the_new.html, @url:https://medium.com/twenty-one-hundred/data-is-the-new-oil-a-ludicrous-proposition-1d91bba4f294],
-Ratner, Bach, and Ré described training data as "The _New_ New Oil" [@url:http://hazyresearch.github.io/snorkel/blog/weak_supervision.html].
-Based on our review of research on deep learning methods to categorize disease,
-this rings true. We expect each of these approaches to play an important role
-if deep learning is going to transform how we analyze data to categorize states
-of human health. Finally, we don't expect that these methods will replace expert
-review. We expect them to complement expert review by allowing more efficient
-use of the costly practice of manual annotation.
+integrated. Numerous writers have described data as the new oil
+[@url:http://ana.blogs.com/maestros/2006/11/data_is_the_new.html, @url:https://medium.com/twenty-one-hundred/data-is-the-new-oil-a-ludicrous-proposition-1d91bba4f294].
+The idea behind this metaphor is that data are available in large quantities,
+valuable once refined, and the underlying resource that will enable a
+data-driven revolution in how work is done. Contrasting with this perspective,
+Ratner, Bach, and Ré described labeled training data as "The _New_ New Oil"
+[@url:http://hazyresearch.github.io/snorkel/blog/weak_supervision.html]. In this
+framing, data are abundant and not a scarce resource. Instead, new approaches
+to solve problems arise when labeled training data become sufficient to enable
+them. Based on our review of research on deep learning methods to categorize
+disease, the latter framing rings true. We expect each of these approaches to
+play an important role if deep learning is going to transform how we analyze
+data to categorize states of human health. Finally, we don't expect that these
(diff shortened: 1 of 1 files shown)


traversc:
Hi Casey, some comments:

>  The new rules are perhaps best summed up by the phrase "right to explanation".   I.e., a clinician must be able to explain to a "right to explanation". 

What I was trying to say, was that the "right to explanation" is only one small part of a the legislation dealing with patient privacy.  So it isn't accurate to say that all the new rules are summed up by that phrase.  

> For example, if we train an algorithm to predict which drugs would be prescribed during a patient's visit to the doctor and there's an existing pattern of racial differences in prescription behavior (TODO: @traversc - you can pick a different example but I think we need one - whichever one you think is most fully supported by the literature.),

Some preliminary literature search didn't turn up much at all, unfortunately.  

However, it also seems a bit speculative, since my understanding is that they're trying to preempt this type of discrimination from happening, not that discriminatory machine learning algorithms are a big problem right now.  Perhaps it would be best to simply stick with summarizing the point from that article?

sections/03_categorize.md
```diff
@@ -196,55 +196,73 @@ this field.
...
+This challenge is not unique to EHR-based studies. Work on medical images,
+-omics data in applications for which detailed metadata are required, and other
+applications for which labels are costly to obtain will be hampered as long as
+abundant curated data are unavailable. Unsupervised and semi-supervised methods
+provide one path forward [@doi:10.1016/j.jbi.2016.10.007], as do adversarial
+training examples [@doi:10.1101/095786]. We also expect the recently described
+data programming strategy [@arxiv:1605.07723] to play a role in addressing this
+challenge. In data programming, noisy automated labeling functions are
+integrated. While numerous writers have described data as the new oil [@url:http://ana.blogs.com/maestros/2006/11/data_is_the_new.html, @url:https://medium.com/twenty-one-hundred/data-is-the-new-oil-a-ludicrous-proposition-1d91bba4f294],
+Ratner, Bach, and Ré described training data as "The _New_ New Oil" [@url:http://hazyresearch.github.io/snorkel/blog/weak_supervision.html].
```
agitter: The phrase "_New_ New Oil" and how it differs from "New Oil" does not make sense without reading the reference.  I suggest explaining that phrase.

sections/03_categorize.md
```diff
@@ -196,55 +196,73 @@ this field.
...
+task of acquiring enough expert-validated labeled data. For instance, in the
+eMERGE consortia and PheKB database [@url:https://phekb.org/implementations],
+most samples with expert validation contain only 100 to 300 patients. These
+datasets are quite small, even for simple machine learning algorithms. The
+challenge is greater for deep learning models with many parameters. While
+unsupervised and semi-supervised approaches can help with small sample sizes,
+the field would benefit greatly from large collections of anonymized records in
+which a substantial number of records have undergone expert review.
+
+This challenge is not unique to EHR-based studies. Work on medical images,
```
agitter: I think we can leave this section as-in, and we'll likely want to refer back to in in our Discussion.  As you say here, data limitations are prevalent in most of the domains we're covering.

sections/03_categorize.md
```diff
@@ -196,55 +196,73 @@ this field.
...
-detect, but difficult for human understanding.  Machine learning and deep
-learning algorithms, which fixate solely on predictive accuracy, may be regarded
-as "black boxes" and are generally not tractable to human understanding.  
-
-Yet the power of machine learning is not disputed.  In the context of medical
-applications, machine learning holds the promise to save more lives and provide
-better quality of life for those it does save.  There is certainly no clear cut
-answer to balance patient rights and utilitarian gain, and these issues will
-only become more prominent as technology advances.  
+[@arxiv:1606.08813v3]. The new rules are perhaps best summed up by the phrase
```
agitter: Will need to account for @traversc's comment before merging.

sections/03_categorize.md
```diff
@@ -196,55 +196,73 @@ this field.
...
+patient who is aided by a machine learning algorithm may be expected to explain
+decisions that use the patient's data.  The new rules were designed to target
+categorization or recommendation systems, which inherently profile individuals,
+but can do so in ways that are discriminatory and unlawful `TODO: @traversc citation needed`.
+
+As datasets become larger and more complex, we may begin to identify
+relationships in data that are important for human health but difficult to
+understand. The algorithms described in this review and others like them may
+become highly accurate and useful for various purposes, including within medical
+practice. However, to discover and avoid discriminatory applications it will be
```
agitter: There is a lot of work and discussion, not necessarily in the medical domain, on discrimination in machine learning.  Do we want to bring in external references?

cgreene:
Merging. Created #302 to help to refine the discussion & revisit this once it's addressed.

PR #298, opened 2017-04-07
Replace numbered citation_id with 4-byte blake2 hash

dhimmel:
Currently, if you click on a reference in the HTML output, you will get directed to the relevant bibliography item.

For example, click on reference 2 and you will go to https://greenelab.github.io/deep-review/#ref-ref_146. This is confusing since `#ref-ref_146` is an internal citation_id that should not be exposed to the end user. This PR replaces numbered citation IDs with random strings (4-byte blake2 hashes) based on their `standard_citation`.

So for example, `#ref-ref_146` becomes `#ref-0ee8b142`. The purpose is to prevent any confusion.

### DIFF
--- a/build/citations.py
+++ b/build/citations.py
@@ -1,3 +1,4 @@
+import base64
 import collections
 from hashlib import blake2b
 import pathlib
@@ -108,8 +109,8 @@ def citation_to_metadata(citation, cache={}):
         msg = f'Unsupported citation  source {source} in {citation}'
         raise ValueError(msg)
 
-    blake2_hash = blake2b(standard_citation.encode(), digest_size=4)
-    citation_id = blake2_hash.hexdigest()
+    digest = blake2b(standard_citation.encode(), digest_size=5).digest()
+    citation_id = base64.b32encode(digest).decode()
     result['citation_id'] = citation_id
     if 'citeproc' in result:
         result['citeproc'] = citeproc_passthrough(
--- a/build/citations.py
+++ b/build/citations.py
@@ -1,4 +1,5 @@
 import collections
+from hashlib import blake2b
 import pathlib
 import re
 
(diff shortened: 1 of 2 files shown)


dhimmel:
I just reset the cache on Travis, to regenerate all `citation_id`.

dhimmel:
Using this [collision calculator](http://davidjohnstone.net/pages/hash-collision-probability), with a 32 bit hash and 1000 references, there is a 0.012% chance of a collision. I'm wondering if we should increase to a 5 or 6 byte hash?

dhimmel:
Switching to base64 encoding instead of hex encoding of the blake2 digest can pack 6 bytes into 8 characters.

dhimmel:
> Switching to base64 encoding instead of hex encoding of the blake2 digest can pack 6 bytes (as opposed to 4) into 8 characters.

URL-safe base64 encoding was messing up pandoc-citeproc due to trailing hyphens. Therefore, in https://github.com/greenelab/deep-review/pull/298/commits/859bb8536a3c9c38bdb2f65e8293f2ad65a68594 I used base32 encoding. Ideally, we'd use base62 encoding to take advantage of all alphanumeric characters, but python 3.6 doesn't support base62. So is it worth a dependency?

agitter:
I did find the integer citation ids a little confusing so this is a good change.  I don't have a strong opinion regarding additional dependencies.  This will mostly run on the Travis servers as far as I can tell (as opposed to different user environments), so if it is easy to set up the environment on Travis I don't see much of a downside to adding dependencies.

Collisions are very unlikely, but would a collision be silent or be reported?

dhimmel:
> I don't see much of a downside to adding dependencies

Agree since it's easy to add to the `environment.yml` and you shouldn't be running this code without that exact environment. The bigger issue is that I couldn't find a good base62 encoding package for python. Several existed but looked abandoned. [`suminb/base62`](https://github.com/suminb/base62) ([PyPI](https://pypi.python.org/pypi/pybase62/)) was one of the few to have a codebase that was still being hosted.

> Collisions are very unlikely, but would a collision be silent or be reported?

Currently, a collision would not be detected and would either cause a downstream error or corrupt the reference that collides. I should explicitly check for collisions and fail if they occur.

dhimmel:
I'm going to merge. I plan to switch from base32 to base62 encoding when `suminb/base62` is ready. Will address in a separate PR.

PR #300, opened 2017-04-08
Discussion placeholder text

agitter:
Adding some reminders about ideas I'd like to cover when we write the discussion in the next week or two.


agitter:
#312 covers LIME and other interpretation methods.  After we merge that, we may decide to move that text to this Discussion section.

PR #303, opened 2017-04-10
Fix indentation bug

dhimmel:

PR #304, opened 2017-04-11
Update DOI of published paper

agitter:
Will merge myself once the build passes

### DIFF
--- a/references/tags.tsv
+++ b/references/tags.tsv
@@ -4,7 +4,7 @@ Alipanahi2015_predicting	doi:10.1038/nmeth.3547
 AltaeTran2016_one_shot	arxiv:1611.03199
 Asgari	doi:10.1371/journal.pone.0141287
 blast	doi:10.1016/S0022-2836(05)80360-2
-Angermueller2016_single_methyl	doi:10.1101/055715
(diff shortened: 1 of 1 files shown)


agitter:
@dhimmel I can't figure out why the build failed with an error about `@tag:Soueidan`.  The DOI in `tags.tsv` is `doi:10.1515/metgen-2016-00011`, and I can resolve that at www.doi.org.

dhimmel:
There's some sort of upstream issue, since

```sh
curl --location \
  --header "Accept: application/vnd.citationstycation/rdf+xml" \
  https://doi.org/10.1515/metgen-2016-0001
```

returns `No acceptable resource available.` See also https://github.com/greenelab/deep-review/pull/303#issuecomment-293071025.

The metadata for that study is showing up [on Crossref](https://search.crossref.org/?q=10.1515%2Fmetgen-2016-0001).

Usually the metadata caching hides these temperamental failures, but I reset the cache for the `citation_id` updates.

agitter:
Thanks for investigating.  Because the issue isn't related to the DOI I changed, I'll merge this.

dhimmel:
> Thanks for investigating. Because the issue isn't related to the DOI I changed, I'll merge this.

Yep that's what we'll have to do until this issue is resolved.

PR #305, opened 2017-04-11
Encode citation_ids using base62

dhimmel:
Builds off of the merged https://github.com/greenelab/deep-review/pull/298.

Thanks @suminb for adding the ability to base62 encode bytes in https://github.com/suminb/base62/pull/6!

I switched to 6-byte hash... now the probability of collision with 1 million references is 0.18%.

 Builds will still fail until the Crossref API issues are fixed https://github.com/CrossRef/rest-api-doc/issues/194

### DIFF
--- a/build/citations.py
+++ b/build/citations.py
@@ -1,9 +1,9 @@
-import base64
 import collections
 from hashlib import blake2b
 import pathlib
 import re
 
+import base62
 import bibtexparser
 
 import metadata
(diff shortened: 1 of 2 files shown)


dhimmel:
@agitter all the build problems have been resolved. The new `citation_id` format is in effect and visible in https://greenelab.github.io/deep-review/.

You can see what the base62 encoded hashes look like in the `citation_id` column of [`processed-citations.tsv`](https://github.com/greenelab/deep-review/blob/a75713eb9036db8258f393d1aa86acc7339af07f/processed-citations.tsv). Quite beautiful.

agitter:
Quite beautiful indeed.  Nice work!

PR #315, opened 2017-04-18
Introduce Categorize Section

cgreene:
I introduce the categorize section with this. I also pruned the only remaining entirely empty subsection.

PR #323, opened 2017-04-20
Move "wide data" to the discussion

cgreene:
This topic is broader (pun semi-intended) than just the categorize section. Therefore, we are going to talk about about it in the discussion section. This is just a routine PR to move it there.


cgreene:
Looks like this contains some commits I didn't intend.

cgreene:
Let's do #315 first and then come back to this one. Should be quick once we get here.

PR #327, opened 2017-04-20
Move interpretability paragraphs

agitter:
I'm moving the text @blengerich wrote in Treat to Discussion so that @akundaje can build on it.

### DIFF
--- a/sections/05_treat.md
+++ b/sections/05_treat.md
@@ -91,46 +91,11 @@ directly assign treatment plans without interpretability; however, the removal
 of human experts from the decision-making loop make the models difficult to
 integrate with clinical practice. To alleviate this challenge, several studies
 have attempted to create more interpretable deep models, either specifically
-for healthcare or as a general procedure for deep learning.
-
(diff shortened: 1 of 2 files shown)


agitter:
@blengerich can you please see if I should move some of this text back to the Treat-specific section?  I didn't have much time right now to carefully divide what you wrote.

agitter:
@blengerich I made a few more minor edits to help the text fit better in the general Discussion section and am now merging this so that others can build on it.  Please open a new pull request if you want to move some of the clinical-specific text back to the paragraphs you wrote in Treat, provide more context here, or otherwise repair some of my quick edits.

blengerich:
Thanks, I agree this fits better in the discussion section. I think there are some subtle differences between interpretation for understanding v.s. interpretation for action, but this should be also included in the discussion section. I will write a short paragraph to clarify these thoughts.

agitter:
Yes, that's an important distinction.  Thanks for offering to clarify my quick edits.

PR #322, opened 2017-04-20
Add opportunities to challenges in categorize

cgreene:
We debated having separate "opportunities" and "challenges" sections; however, as I was writing these it became clear that they kept referring to parallel parts of each other. I've now integrated the text from both sections and removed the separate opportunities section.

@brettbj - I think this hit your subsection most strongly so I'd consider you the primary reviewer. @agitter - feel free to take a look if you want to. I realize that you are also quite busy, so if @brettbj approves the changes my inclination is to merge.

PR #328, opened 2017-04-21
(categorize/EHR): add federated DL and secure aggregations

alxndrkalinin:
- remove duplicate `Data sharing and privacy` section
- add federated DL and secure aggregations


brettbj:
@cgreene made fairly major changes to this section. I think this is worth including but might want to merge into #322

alxndrkalinin:
@brettbj thanks, completely missed that, will merge to #322

PR #334, opened 2017-04-21
Fix Tags in Treat Section

blengerich:
Transfer from in-line doi or arxiv references to internal tags.


agitter:
Look good, I'll merge.  I like to use tags because it helps me when editing my draft text but don't feel compelled to use them if you prefer not to.

blengerich:
Thanks, I also prefer the tags for readability. Had a few minutes today, so I also switched the references in the introduction section to use internal tags (#334).

PR #246, opened 2017-02-17
first draft of intro addition

swamidass:
I think this has my key points. Curious everyone's thoughts. 

It does need to have references added.

PR #335, opened 2017-04-21
Switch to tags in Intro Section

blengerich:
Transfer from in-line doi or arxiv references to internal tags.


dhimmel:
@agitter and @blengerich the reason I prefer inline references to tags is that tags introduce a lot more room for error. In order for a tag reference to work, you must do multiple things correctly (across multiple files). The difficulty in doing multiple things correctly is evidenced by the commit history of this PR.

Additionally tags create a lookup burden for anyone who's not familiar with the study. You can't immediately go to the cited work, you have to look it up in `tags.tsv`. Also with multiple people editing `tags.tsv` at once, conflicts will become more frequent.

Not that we shouldn't merge this PR, but I caution you from going too far down the tag rabbit hole.

blengerich:
Your points make sense, especially with the errors in this PR commit history 😄. The choice of style doesn't make a big difference to me - I just had a few minutes and figured I could help out a little in case we did end up going with the tag style. Maintainers, please feel free to close this PR if desired.

agitter:
Should we be open-ended about tags versus inline for individual sub-sections, which are primarily written and edited by one person, but not encourage tags globally?

I'm hesitant to edit the intro in particular right now because we need to resolve #246.  I just noticed to converting to tags in Treat introduced conflicts in my #313 so we may want to hold off further widespread tag conversion.

blengerich:
Sounds reasonable. I will close this PR and it can be reopened later if desired.

dhimmel:
> Should we be open-ended about tags versus inline for individual sub-sections, which are primarily written and edited by one person, but not encourage tags globally?

I'm inclined to let authors do what they like... but we shouldn't give off the impression that tags are the preferred approach. I think tags make sense when the reference is unwieldy, like an ugly URL or long disgusting DOI that has no human readable portions. Or potentially if you're repeatedly referencing the same thing. But in general, the tag is more trouble than its worth (IMO).

PR #344, opened 2017-04-23
initial draft of sequencing and variant calling

bdo311:
Thanks in advance for the feedback! I'll probably try to revise some of the wording to improve the flow.

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -605,5 +605,65 @@ unlabeled (microbiome experiments) examples.
 
 ### Sequencing and variant calling
 
-*We have one nanopore paper in the issues and very recent work on variant calling
-that looks worthy of inclusion.*
+While we have so far discussed the role of deep learning in analyzing genomic
+data, deep learning approaches can also substantially improve our ability to
+obtain the genomic data itself. We will discuss two specific challenges: calling
+SNPs (single nucleotide polymorphisms) and indels (insertions and deletions)
+with high specificity and sensitivity, and improving the accuracy of new types
+of data such as nanopore sequencing. These two tasks are critical for studying
+rare variation, allele-specific transcription and translation, and splice site
+mutations, among others. In the clinical realm, sequencing of rare tumor clones
+and other genetic diseases will require accurate calling of SNP and indels.
+
+Current methods achieve relatively high (>99%) precision at 90% recall for SNPs
+and indel calls from Illumina short-read data, yet this leaves a large number of
+potentially clinically significant remaining false positives and false
+negatives. These methods have so far relied on experts to build probabilistic
+models that reliably separate signal from noise. However, this process is time
+consuming and, more importantly, fundamentally limited by how well we understand
+and can model the factors that contribute to noise. Recently, two groups have
+applied deep learning to construct data-driven and, therefore, unbiased noise
+models. One of these models, DeepVariant, builds directly on well-established
+deep learning methods for image recognition by encoding reads around a candidate
+SNP as a 221x100 bitmap image, where each column is a nucleotide and each row is
+a read from the sample library [@tag:Poplin2016_deepvariant]. The top 5 rows
+represent the reference, and the bottom 95 rows represent randomly sampled reads
+that overlap the candidate variant. Each RGBA (red/green/blue/alpha) image pixel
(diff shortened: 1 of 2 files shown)


sections/04_study.md
```diff
@@ -605,5 +605,65 @@ unlabeled (microbiome experiments) examples.
...
+obtain the genomic data itself. We will discuss two specific challenges: calling
+SNPs (single nucleotide polymorphisms) and indels (insertions and deletions)
+with high specificity and sensitivity, and improving the accuracy of new types
+of data such as nanopore sequencing. These two tasks are critical for studying
+rare variation, allele-specific transcription and translation, and splice site
+mutations, among others. In the clinical realm, sequencing of rare tumor clones
+and other genetic diseases will require accurate calling of SNP and indels.
+
+Current methods achieve relatively high (>99%) precision at 90% recall for SNPs
+and indel calls from Illumina short-read data, yet this leaves a large number of
```
agitter: Do we have a reference we can use for this performance level?

sections/04_study.md
```diff
@@ -605,5 +605,65 @@ unlabeled (microbiome experiments) examples.
...
+encodes the base (A, C, T, G) as a different R value, quality score as a G
+value, strand as a B value, and variation from the reference as the alpha value.
+The neural network outputs genotype probabilities for each candidate variant.
+They were able to achieve better performance than GATK, a leading genotype
+caller, even when GATK was given information about population variation for each
+candidate variant. Another method, still in its infancy, hand-developed 642
+features for each candidate variant and fed these vectors into a fully connected
+deep neural network [@tag:Torracinta2016_deep_snp]. Unfortunately, this feature
+set required at least 15 iterations of software development to fine-tune, which
+will likely not be generalizable. Going forward, we foresee that variant calling
```
agitter: Great conclusions, I'm always happy to see us take a stance in the review.  To challenge you slightly, what do you think about the tension between natural encodings of the data where domain-specific labeled data may be limited versus conceptually suboptimal encodings (e.g. DNA -> images) where labeled data are rich or pre-trained models are available?  Is this image-to-variant calling strategy likely to hold an advantage only in the short term until better training datasets or simulators become available?  Should it be more widely adopted elsewhere in genomics?

sections/04_study.md
```diff
@@ -605,5 +605,65 @@ unlabeled (microbiome experiments) examples.
...
+will benefit most from optimized neural network architectures and better methods
+of encoding raw sequence and quality data (e.g. in tensor rather than RGBA
+format).
+
+In limited experiments, DeepVariant was robust to sequencing depth, read length,
+and even species [@tag:Poplin2016_deepvariant]. However, a model built on
+Illumina data, for instance, will likely not be applicable to PacBio long-read
+data or MinION nanopore data, which have vastly different specificity and
+sensitivity profiles and signal-to-noise characteristics. Recently, Boza et al.
+used bidirectional recurrent neural networks infer the E. coli sequence from
```
agitter: _E. coli_ in italics

sections/04_study.md
bdo311: Yep I'll re-reference the Poplin paper, which benchmarked the previous state of the art algorithms (GATK).

sections/04_study.md
bdo311: I think it's an issue of not really knowing the best representation with which to input your data. With sequence data, we could either encode it as an RGB image and send it through a CNN already pretrained on ImageNet, or encode it as a one-hot tensor and send it through a CNN that we train and optimize explicit for genomic data. To know which one is better will require more work to be done. 

I think what I was trying to get across in the section was that inputting raw sequence (in its optimal representation) could probably in the long run be superior or equal to inputting sequence + hand-derived features (in its optimal representation) in terms of time spent optimizing.

agitter: The point about raw sequence versus hand-derived features came across clearly.  If we think comparing image-based versus natural encodings will require more work to be done, I'll merge this as-is.

bdo311: I can try to clarify the point about image based vs natural encodings in a sentence -- just to say that we don't know which will be best, just that there needs to be more research to test both strategies out under otherwise identical conditions.

agitter: Okay.  You only need to add that if you really want to.  I merged already so you would need to open another small pull request.

Issue #348, opened 2017-04-24
Move data sharing paragraph


agitter:
In planning the "Code, data, and model sharing" sub-section of Discussion, I realized it would be helpful to move the "research parasites" text here.  @cgreene if you agree can you please move the Categorize paragraph starting with "In addition to methodological improvements" to Discussion?  I can do it myself but thought your Categorize section might also need to be modified to link the data sharing discussion there with the relevant Discussion sub-section.

agapow:
This may have been reintroduced with my merge on the categorize section. Mea culpa. I can rip it out if need be.

PR #350, opened 2017-04-24
Address #348 by moving text to discussion.

cgreene:
This moves the paragraph from #348 to the discussion. I also lightly re-worked what's there. On our more substantial revisions we'll probably need to play around with flow, since there's a separate "Data sharing" section that comes immediately after this. I'll play around with that a little bit more to see if I can keep the message that culture is important while also touching on the technical challenges within this part.

### DIFF
--- a/sections/03_categorize.md
+++ b/sections/03_categorize.md
@@ -270,23 +270,14 @@ to solve problems arise when labeled training data become sufficient to enable
 them. Based on our review of research on deep learning methods to categorize
 disease, the latter framing rings true.
 
-In addition to methodological improvements, a robust culture of data sharing -
-and in particular the sharing of high-quality labeled datasets - would do much
-to speed advances in this domain. The cultural barriers are perhaps best
-captured by the implications of using the term "research parasite" to describe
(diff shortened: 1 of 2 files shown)


cgreene:
Ok - @agitter - see what you think on this. I was trying to address #348 & define what we discuss in categorize vs punt to discussion.

sections/03_categorize.md
```diff
@@ -270,26 +270,21 @@ to solve problems arise when labeled training data become sufficient to enable
...
+health. We don't expect that deep learning methods will replace expert review.
+We expect them to complement expert review by allowing more efficient use of the
+costly practice of manual annotation.
 
 ###### Data sharing is hampered by standardization and privacy considerations
 
+To construct the types of very large datasets that deep learning methods thrive
+on, we need robust sharing of large collections of data. This is in part a
+cultural challenge. We touch on this challenge in the Discussion section.
+However, even if addressed the cultural hurdles around data sharing there would
```
agitter: Missing word in this line?

agitter:
Looks good

PR #355, opened 2017-04-25
draft of miRNA section

bdo311:
I tried to keep this section compact. Thanks in advance for the feedback!

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -271,8 +271,8 @@ great interest, as they are critical components of gene regulatory networks and
 are often conserved across great evolutionary distance [@tag:Bracken2016_mirna
 @tag:Berezikov2011_mirna]. While many machine learning algorithms have been
 applied to solve these prediction tasks, they currently require extensive
-feature selection and optimization. For instance, the widely adopted state of
-the art for miRNA target prediction, TargetScan, trained multiple linear
+feature selection and optimization. For instance, one of the most widely adopted
+tools for miRNA target prediction, TargetScan, trained multiple linear
 regression models on 14 hand-curated features including structural accessibility
 of the target site on the mRNA, the degree of site conservation, and predicted
 thermodynamic stability of the miRNA:mRNA complex [@tag:Agarwal2015_targetscan].
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -266,8 +266,40 @@ application in this area.*
 
 ### Micro-RNA binding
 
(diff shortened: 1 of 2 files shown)


brettbj:
@bdo311 others will be better positioned to do a full review, but could you change your tags to either [@doi:link] or [@arxiv:link] where available? There is a script to automatically process citations in this format (https://github.com/greenelab/deep-review/blob/master/CONTRIBUTING.md#markdown)

bdo311:
@brettbj I've been following the instructions to use descriptive tags for the references. Is that something we're not doing anymore?

"For improved readability, you can (but are not required to) use a tag. For example, you can add a reference to the tag Zhou2015_deep_sea using the following syntax: [@tag:Zhou2015_deep_sea]. If you add references that use tags, make sure to add those tags and their corresponding citations to references/tags.tsv."

brettbj:
@bdo311 my apologies, I didn't see that had been added as an option, looks good then, will leave for someone with more miRNA knowledge

sections/04_study.md
```diff
@@ -266,8 +266,40 @@ application in this area.*
...
 ### Micro-RNA binding
 
-*miRNAs are important biologically, but have neural networks produced anything
-particularly notable in this area?*
+Prediction of microRNAs (miRNAs) in the genome as well as miRNA targets is of
+great interest, as they are critical components of gene regulatory networks and
+are often conserved across great evolutionary distance [@tag:Bracken2016_mirna
+@tag:Berezikov2011_mirna]. While many machine learning algorithms have been
+applied to solve these prediction tasks, they currently require extensive
+feature selection and optimization. For instance, the widely adopted state of
```
agitter: I don't have any evidence to the contrary, but do you have a sense of how accepted it is that TargetScan is the state of the art?  There are other competitors, but I do tend to see TargetScan used frequently.

I'm wondering whether we need to hedge slightly and say TargetScan is an example of state of the art versus it is *the* state of the art.  If it is clearly the dominant existing method, I prefer to keep the text as-is.

bdo311:
I was thinking about this as well @agitter -- not sure what super-accurate miRNA target prediction would provide, especially because the applications right now are more in basic biology than in therapeutics, where siRNAs and gapmers are preferred. Would love to hear thoughts from others who know the field better.

agitter:
I don't have any specific ideas about the consequences of improved miRNA binding predictions either.  I'll merge this soon.  If someone wants to add something during our editing period they should feel welcome to do so.

PR #347, opened 2017-04-24
Discussion: transfer learning

alxndrkalinin:
Here is initial draft for [transfer learning and multimodal learning section](https://github.com/greenelab/deep-review/issues/188#issuecomment-296204170).
Feedback for clarity, accessibility, structure, word flow, and grammar would be highly appreciated. I will have time to revise later this week. If I missed any papers, please also let me know and I will try my best to include them.

PR #356, opened 2017-04-25
add TFBS todos

jacklanchantin:


agitter:
I see two tags in a `TODO`.  Can you please add those to `tags.tsv` so we don't get out of sync across pull requests?  Or you can skip the tags and use the DOIs directly.

agitter:
@agapow after reading this, do you see a reasonable path forward merging this and the promoter/enhancer section you wrote?  I'm open to the idea but haven't thought about the structure of the merged section or the benefits of combining them versus keeping them separate with the latter referring to the former.

agapow:
There's a lot of common elements (ChipSeq, CNNs, training) and identifying enhancers can be seen as a subset of identifying TFBSs. Maybe a section introducing the broad area of identifying regulatory sites, followed by discussion of particular papers, end with summary of opportunities / problems.

Alternatively, it may be that the promoter section refers back to the TFBS ("here too we see issues with ...")

sections/04_study.md
```diff
@@ -222,13 +222,67 @@ systems are limited by expensive computational cost proportional to the number
...
+Deep learning models have shown great accuracy on the TFBS task, but
+the results are not fully convincing for several reasons based on the
+datasets used and the evaluation of the model's predictions on such
+datasets. First, ChIP-seq experiments give a continuous value of binding 
+likelihood at a certain location based on many experiments at that
+location in DNA. Based on these values, the TFBS task is usually
+converted into a binary classification task based on a certain threshold
+of the ChIP-seq value. However, the value may not be accurate in itself.
+There is no guarantee of a binding or non binding at that location
+depending on which person's sequence it is. Second, most datasets predict
```
agitter: Most classifiers?  Or is the point that ChIP-seq tags a single TF at a time?

sections/04_study.md
```diff
@@ -222,13 +222,67 @@ systems are limited by expensive computational cost proportional to the number
...
+one TF at a time, using a separate model for each TF. In reality, there
```
agitter: Isn't a lot of work in this domain multi-task? Or are you referring to something else?

sections/04_study.md
```diff
@@ -222,13 +222,67 @@ systems are limited by expensive computational cost proportional to the number
...
+may be multiple TFs binding at the same location and interacting, thus
+requiring both a dataset which gives all TF binding values at every location
+and a multi-task model. Third, it is unclear exactly
```
agitter: Is this related to training on class-skewed data?  It seems like the suggestion is to subsample the negative sites during training.  Is that better than using class weights, or am I missing the point?

sections/04_study.md
```diff
@@ -222,13 +222,67 @@ systems are limited by expensive computational cost proportional to the number
...
+These dataset formulations results in inaccuracies of the evaluation metrics.
+Converting the task into a binary classification isn't completely accurate
+since the model may give a high probability of a binding site, but the
+ChIP-seq binding signal may be just barely over the binary classification
+threshold, resulting in potential false positive (or the signal may be just
+below the threshold, resulting in potential false negatives). One of the 
+biggest flaws is the fact that most datasets artificially balance the
+positive and negative binding binding sites and report auROC for the metric.
+This is very misleading in a task where the binding sites are very unevenly
+balanced in the real world. Thus, we need datasets which more accurately
```
agitter: What specifically?  Are existing continuous values provided in ChIP-seq insufficient, and if so what would be more useful for training neural nets?  Or is it just that the data aren't being used properly?

sections/04_study.md
```diff
@@ -222,13 +222,67 @@ systems are limited by expensive computational cost proportional to the number
...
+
+These dataset formulations results in inaccuracies of the evaluation metrics.
+Converting the task into a binary classification isn't completely accurate
+since the model may give a high probability of a binding site, but the
+ChIP-seq binding signal may be just barely over the binary classification
+threshold, resulting in potential false positive (or the signal may be just
+below the threshold, resulting in potential false negatives). One of the 
+biggest flaws is the fact that most datasets artificially balance the
+positive and negative binding binding sites and report auROC for the metric.
+This is very misleading in a task where the binding sites are very unevenly
```
agitter: After this sentence, you could add "balanced in the real world (see Discussion)" because we'll add an evaluation section there.

sections/04_study.md
agapow: This would refer to the imbalanced nature of "real" data, i.e. that there's much many more negative cases than positive cases (more sites that don't bind anything or don't bind a given ligand / protein). So there's an issue on how to appropriately generate and generate training data and how to assess performance.

agitter: My confusion was with "Thus, we need datasets which more accurately model real TFBS data."  Does this refer to real vs. continuous labels, skewed positive vs negative instances, or something else?  A ChIP-seq dataset will already have few peaks relative to the number of possible binding sites (i.e. show class skew if we binarize it), so is there a different type of training data we would like to see?

agitter:
Note that #347 on transfer learning now discusses multi-task TF binding methods.  We don't need to repeat that discussion here, but we should refer to it somewhere in this sub-section.

sections/04_study.md
```diff
@@ -222,13 +222,67 @@ systems are limited by expensive computational cost proportional to the number
...
-samples and TFs. While neural architectures are rapidly changing and producing
+to a large number of genomic sequences. 
+
+Since the work by [@tag:Alipanahi2015_predicting] was published, there have
+been a multitude of deep learning works on the TFBS task. Due to the 
+motif-driven nature of the TFBS task, most architectures
+have been convolutional-based, as summarized in [@tag:Zeng2016_convolutional].
+[@tag:Lanchantin2016_motif] introduced several new convolutional and 
+recurrent neural network models for predicting TFBSs, which showed improvements
+on the baseline datasets. It is also important
```
jacklanchantin: Deepbind dataset. Updated that.

sections/04_study.md
jacklanchantin: Should be:

"The second issue is that most datasets are split by TF,
requiring a separate model for each TF (i.e. binary classification
on each TF sub-dataset)."

sections/04_study.md
jacklanchantin: Most of the papers I know about use datasets where it is only predicting one TF binding at a time (as opposed to all TFs at once). I don't know of any datasets that have all TF bindings for a particular position. This is actually something I'm currently generating for a new project.

Do you have a particular example?

sections/04_study.md
jacklanchantin: Is your question for the "second issue"?

I think the class-skewed data is brought up in the "third issue"

akundaje:
The DeepSEA paper learns a multitask model across multiple TFs. But it is
sequence only so doesn't generalize across cell types.

Anshul

jacklanchantin:
That is true, thanks for the input. I actually haven't mentioned the
DeepSEA paper in the TFBS section since TF binding prediction is an
intermediate step in their model and it isn't heavily analyzed, but maybe I
should talk about it?

akundaje:
I think it's worth mentioning.

Btw, one massive fatal flaw in all sequence only TF models is they cannot
by definition predict binding in new unseen cell types/conditions. So in
some sense they are useless for "prediction". You have to include another
data mode that changes across cell types/conditions like chromatin
accessibility/marks, RNA-SEQ, DNA methylation.

Anshul



Anshul

agitter:
@akundaje I totally agree that this is important to bring up.  That's my intention with the TODO comment about #258, the recent TFImpute paper in PLoS Comp Bio, which we need to incorporate here.  Until the DREAM methods come out, they are the only deep learning method I'm aware of that attempts prediction in new cell type-TF pairs.  But their cell types (conditions) still have to be in the training dataset.  It's not as fully general as what you describe.

@jacklanchantin I also recommend including DeepSEA, with the context @akundaje raised, because they do predict ChIP-seq for some of their tasks.

jacklanchantin:
@akundaje <https://github.com/akundaje> That's a really good point. I
should've mentioned that earlier. I will add both of those points in. Thank
you!

jacklanchantin:
@akundaje <https://github.com/akundaje> @agitter
<https://github.com/agitter> I think that this is one of the main issues in
general in tasks on data from repositories such as ENCODE. Most works just
show that the models can extract the right features to understand the task
(by making high accuracy predictions on a particular cell type). But as
Anshul noted, this is fairly useless in the biomedical world since we have
data such as ChIP-seq for many cell types. So I think we should talk about
domain transfer (e.g. from one cell type to another), but I think this
extends past just TFBS prediction, so I'm not sure where to put it.

sections/04_study.md
jacklanchantin: Added

akundaje:
Even without domain transfer one can still train multimodal models on
sequence + one or more cell-type specific modes. Everyone in the DREAM
challenge did this. Noone has really ventured in domain transfer for TF
binding except for that one paper from your/Yanjun's lab.

So I think in the TF section you can discuss how multimodal models are
absolutely necessary for cross sample prediction (which are lacking in the
current literature) which is the realistic application for a predictive
model. E.g of such modalities would be chromatin accessibility, DNA
methylation, gene expression which are readily available.

You can briefly mention the additional complication that domain adaptation
methods will also be imperative to account for the fact that the same TF
can have very different cofactors across conditions/cell types. Hence,
sequence features that optimal in one condition may result in very poor
generalization in a new condition. This area of research will be
challenging but exciting.

sections/04_study.md
jacklanchantin: I think it's about both issues. In my experience with talking to people at the UVA med school, they don't think it should be a binary classification problem since the ChIP-seq threshold isn't clear. Either way we formulate the ChIP-seq peak value, the class skew is relevant though. There are only a small number of regions that have ChIP-seq values above a noticeable value.

jacklanchantin: We have been hesitant to make it a regression model to predict ChIP-seq value because that is probably overanalyzing the problem, but binary classification may be under analyzing.

akundaje:
The way we designed it in DREAM was to have a binary classification problem
with an ambiguous label zone. i.e. we had high confidence positives, high
confidence negatives and then regions that are ambiguous where are not
sure. Such examples are not used in training or downweighted. Also not
directly used in test accuracy evaluation. So there is barrier of
uncertainty between what is called positive and negative. So if a
prediction just goes below or above these threshold they don't suddenly
pass into the the wrong class. There are also several other ways to get
around predicting label uncertainty e.g. using predictors that can also
decide to abstain when they are uncertain. One can also treat it as a
binary classification problem with weighted data where the weights come
from the binding intensity values. You can also evaluate the problem where
you focus more on the ranking (auPRC, precision and recall will take this
into account).

Binary classification is not necessarily the wrong formulation. Even though
bulk ChIP-seq appears continuous it is actually measuring the probability
of a binary binding event across a population of cells (unlike gene
expression for example which is truly continuous).

Classical vanilla regression is also not appropriate because the true
ChIP-seq scores potentially matter only in the signal regime i.e in what
would be considered real binding sites. The rest of the genome is just
noise and you dont want to be regressing on noise. Of course this can be
modeled as a mixture of distributions but regression or ranking loss really
matters only in the signal regime.

-Anshul.

agitter:
@jacklanchantin does your commit a868df5 include all the changes you wanted to make?  If so, I'll re-review this and see if we can merge.

@akundaje those are important topics your raise that we might want to work into our discussion section somewhere.  They potentially fit in the "data limitations" sub-section.

jacklanchantin:
@agitter yes this is everything I wanted to include so far

@akundaje I agree. Do you all think that I harped on the issues too much rather than focusing on the review of what's been done, or is this ok?

sections/04_study.md
agitter: @akundaje's comments below are insightful regarding classification versus regression and consistent with what you say here.

agitter:
@jacklanchantin thanks for expanding this section even further.  I'm going to merge this and then go slightly outside of our typical review process to create and merge a new branch that fixes the Travis CI build errors.  I don't have permission to push directly to your branch.

@akundaje and @jisraeli if you have more revisions to make to this section, let's take them to a new pull request.

Issue #342, opened 2017-04-23
Large-Scale Discovery of Disease-Disease and Disease-Gene Associations


enricoferrero:
http://doi.org/10.1038/srep32404 (https://www.nature.com/articles/srep32404)

> Data-driven phenotype analyses on Electronic Health Record (EHR) data have recently drawn benefits across many areas of clinical practice, uncovering new links in the medical sciences that can potentially affect the well-being of millions of patients. In this paper, EHR data is used to discover novel relationships between diseases by studying their comorbidities (co-occurrences in patients). A novel embedding model is designed to extract knowledge from disease comorbidities by learning from a large-scale EHR database comprising more than 35 million inpatient cases spanning nearly a decade, revealing significant improvements on disease phenotyping over current computational approaches. In addition, the use of the proposed methodology is extended to discover novel disease-gene associations by including valuable domain knowledge from genome-wide association studies. To evaluate our approach, its effectiveness is compared against a held-out set where, again, it revealed very compelling results. For selected diseases, we further identify candidate gene lists for which disease-gene associations were not studied previously. Thus, our approach provides biomedical researchers with new tools to filter genes of interest, thus, reducing costly lab studies.

This one is quite interesting and I haven't seen it mentioned in here. It's not overly obvious but my understanding is that they are using deep learning:
> The neural network is trained by projecting the vectors for context words  into a latent representation with **multiple non-linear hidden layers** and the output softmax layer comprising W nodes, where W is a size of the vocabulary (equal to the number of diseases  in our task), while attempting to predict word wt with high probability.

It could fit in the EHR section even though it's not really about categorisation. I think the most interesting bit is about novel gene - disease associations (which could be targets) so maybe Treat is a better place?

agitter:
It might fit in Study or Categorize.  Most of the EHR work has been discussed in Categorize.  The gene-disease associations seem more like Study than Treat to me, though the lines are blurry between these sections.

enricoferrero:
OK, I've just finished reading it and, despite being a methodological paper, the architecture of the (deep?) neural network for their neural embedding model is not very well described in my opinion. Generally speaking the, neural network aspect is not emphasised so I'm not sure it should be included. It's still a cool paper though and the  DAG2D performs very well and could be used effectively for novel target identification.

If @agitter or @brettbj could have a look and advise whether this should be included or not, that'd be great.

[...]

Issue #63, opened 2016-08-06
Deep Patient: An Unsupervised Representation to Predict the Future of Patients from the Electronic Health Records


agitter:
http://doi.org/10.1038/srep26094

cgreene:
Cross referencing to #25 . Both should be discussed together as they are essentially concurrent applications. These do seem relevant to the topic. @brettbj willing to share thoughts on this? I know you have read them in depth. Might want to see if Joel or someone from his lab would contribute thoughts on #25. Clearly both groups are thinking about this.

XieConnect:
@brettbj  If Brett is busy with other tasks, I may help present a brief summary of this paper, and maybe also contrast several other EHR phenotyping and patient subtype works. Mine will be much shorter though, and restrict to their relevance to the survey.

cgreene:
@XieConnect - sounds great! @brettbj : what do you think?

brettbj:
This slipped down the todolist, @XieConnect go for it, I'll add any thoughts I have to your summary

XieConnect:
TL;DR: 3-layer stack of denoising autoencoders were applied on EHR data to learn patient representations. And random forest classifier was trained for disease prediction on holdout dataset. 


## Goal: 
Use unsupervised deep learning to learn better patient representations. This will the help improve subsequent tasks, such as disease prediction (classification) using standard machine learning models. 


## Method: 
3-layer stack of denoising autoencoders were used in the initial unsupervised training stage. 
Random forest classifier was subsequently used in the last step of disease prediction. 


Hyperparameter tuning for autoencoders was performed using a separate validation dataset (tuned by looking at the supervised classification performance). 


## Hyperparameters: 
3 layers stack of denoising autoencoders (more layers have been reported, with no obvious improvement). 
500 hidden units per layer. 
5% noise corruption factor (used zero-out masking noise). 
Sigmoid activation function. 
100 random forest trees. 




## Data: 
EHR data from Mount Sinai. Filtered by thresholding on ICD9 codes (frequencies). 
Patient features include demographics, structured clinical descriptors (ICD9), medication, procedures, lab tests, and text mining (NLP) features (mainly topics from topic modeling). 
Patient disease labels were obtained by simple disease categorization from ICD9 codes, representing 78 diseases after filtering. 


Data in Year 2014 were retained as testing, and data before that as training. 


Final training data: 704,587 patients, and 41,072 features. 




## Experiments


### Data description:  
[...]

PR #357, opened 2017-04-25
Suggested additions/changes to Categorize/EHR

enricoferrero:
OK, here's an attempt at incorporating #342 in the EHR section of Categorize. I think it can stand as its own paragraph, or it can be somehow amalgamated with the previous one (note that this was only very slightly edited).

I would also like to propose a few changes to the description of #63. Please feel free to disagree and revert back to the original :smiley:

Feedback very welcome!

### DIFF
--- a/sections/03_categorize.md
+++ b/sections/03_categorize.md
@@ -181,12 +181,13 @@ showed that unsupervised feature construction of many features via denoising
 autoencoder neural networks could dramatically reduce the number of labeled
 examples required for subsequent supervised analyses
 [@doi:10.1016/j.jbi.2016.10.007]. In addition, it pointed towards learned
-features being useful for subtyping within a single disease. A concurrent large-
-scale analysis of an electronic health records system found that a deep
-denoising autoencoder architecture applied to the number and co-occurrence of
-clinical test events, though not the results of those tests, constructed
-features that were more useful for disease prediction than other existing
-feature construction methods [@doi:10.1038/srep26094].  Razavian et al.
+features being useful for subtyping within a single disease. In a concurrent
+large-scale analysis of EHR data from 7000,000 patients, Miotto et al
+[@doi:10.1038/srep26094] used a deep denoising autoencoder architecture
+applied to the number and co-occurrence of clinical events ("DeepPatient") to
+learn a representation of patients. The model was able to predict disease
+trajectories within one year with over 90% accuracy and patient-level
+predictions were improved by up to 15% when compared to other methods. Razavian et al.
 [@arxiv:1608.00647] used a set of 18 common lab tests to predict disease onset
 using both CNN and LSTM architectures and demonstrated and improvement over baseline
 regression models. However, numerous challenges including data integration (patient
--- a/sections/03_categorize.md
+++ b/sections/03_categorize.md
@@ -142,21 +142,34 @@ This indicates a potential strength of deep methods. It may be possible to
 repurpose features from task to task, improving overall predictions as the field
 tackles new challenges.
 
-Several authors have created reusable feature sets for medical terminologies using
-neural embeddings, as popularized by word2Vec [@tag:Word2Vec]. This approach
-was first used on free text medical notes by De Vine et al.
-[@doi:10.1145/2661829.2661974] with results at or better than traditional methods.
-Y. Choi et al.[@doi:10.1145/2567948.2577348] built embeddings of standardized
-terminologies, such as ICD and NDC, used in widely available administrative
-claims data. By learning terminologies for different entities in the same
-vector space, they can potentially find relationships between different
-domains (e.g. drugs and the diseases they treat). Medical claims data does not
-have the natural document structure of clinical notes, and this issue was
-addressed by E. Choi et al. [@doi:10.1145/2939672.2939823], who built
-embeddings using a multi-layer network architecture which mimics the structure
-of claims data. While promising, difficulties in evaluating the quality of
-these kinds of features and variations in clinical coding practices remain as
-challenges to using them in practice.
+Several authors have created reusable feature sets for medical terminologies
+using natural language processing (NLP) and neural embedding models, as
+popularized by Word2vec [@tag:Word2Vec]. This approach was first used on free
+text medical notes by De Vine et al. [@doi:10.1145/2661829.2661974]
+with results at or better than traditional methods. Y. Choi et al.
+[@doi:10.1145/2567948.2577348] built embeddings of standardized terminologies,
+such as ICD and NDC, used in widely available administrative claims data. By
+learning terminologies for different entities in the same vector space, they can
+potentially find relationships between different domains (e.g. drugs and the
+diseases they treat). Medical claims data does not have the natural document
+structure of clinical notes, and this issue was addressed by E. Choi et al.
+[@doi:10.1145/2939672.2939823], who built embeddings using a multi-layer network
+architecture which mimics the structure of claims data. While promising,
+difficulties in evaluating the quality of these kinds of features and variations
+in clinical coding practices remain as challenges to using them in practice.
+
+NLP has also been applied directly to EHRs to predict disease comorbidities and
+phenotypes as well as novel gene - disease associations. Gligorijevic et al
+[@doi:10.1038/srep32404] built neural embedding models and mined more than 35
+million records collected over a decade. The learned disease representations
+significantly ouperformed state-of-the-art methods, achieving close to 86%
+accuracy. In a second model, the authors supplement patient discharge records
+with genetic diseases associations from genome-wide association studies (GWASs)
+to learn from both disease and gene vectors at the same time, with similar
+perfomance gains. On a held-out set the model recovered all 185 genes known to
+be associated with congestive heart failure and predicted 3 novel genes for
+chronic airway obstruction, showing that mining of EHR data can potentially
+impact the discovery of therapeutic targets.
 
 Identifying consistent subgroups of individuals and individual health
 trajectories from clinical tests is also an active area of research. Approaches
@@ -202,8 +215,8 @@ family distributions [@arxiv:1411.2581v1]. The result was a deep survival
 analysis model capable of overcoming challenges posed by missing data and
 heterogeneous data types, while uncovering nonlinear relationships between
 covariates and failure time. They showed their model more accurately
-stratified patients as a function of disease risk score compared the current
-clinical implementation.
+stratified patients as a function of disease risk score compared to the current
+GWASs clinical implementation.
 
 There is a computational cost for these methods, however, when compared to
 traditional, non-network approaches. For the exponential family models,
--- a/sections/03_categorize.md
+++ b/sections/03_categorize.md
@@ -217,7 +217,7 @@ analysis model capable of overcoming challenges posed by missing data and
 heterogeneous data types, while uncovering nonlinear relationships between
 covariates and failure time. They showed their model more accurately
 stratified patients as a function of disease risk score compared to the current
-GWASs clinical implementation.
+clinical implementation.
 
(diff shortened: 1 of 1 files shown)


sections/03_categorize.md
```diff
@@ -168,19 +181,20 @@ showed that unsupervised feature construction of many features via denoising
...
-feature construction methods [@doi:10.1038/srep26094].  Razavian et al.
-[@arxiv:1608.00647] used a set of 18 common lab tests to predict disease onset
-using both CNN and LSTM architectures and demonstrated and improvement over baseline
-regression models. However, numerous challenges including data integration (patient
-demographics, family history, laboratory tests, text-based patient records,
-image analysis, genomic data) and better handling of streaming temporal data
-with many features, will need to be overcome before we can fully assess the
-potential of deep learning for this application area.
+features being useful for subtyping within a single disease. In a concurrent
+large-scale analysis of EHR data from 7000,000 patients, Miotto et al
```
cgreene: I think it's 700,000

cgreene:
A bit swamped today. Made just a quick scan & think the number is 700k if I recall correctly.

sections/03_categorize.md
enricoferrero: Yes :smile:

sections/03_categorize.md
```diff
@@ -142,21 +142,34 @@ This indicates a potential strength of deep methods. It may be possible to
...
+[@doi:10.1145/2939672.2939823], who built embeddings using a multi-layer network
+architecture which mimics the structure of claims data. While promising,
+difficulties in evaluating the quality of these kinds of features and variations
+in clinical coding practices remain as challenges to using them in practice.
+
+NLP has also been applied directly to EHRs to predict disease comorbidities and
+phenotypes as well as novel gene - disease associations. Gligorijevic et al
+[@doi:10.1038/srep32404] built neural embedding models and mined more than 35
+million records collected over a decade. The learned disease representations
+significantly ouperformed state-of-the-art methods, achieving close to 86%
```
agitter: This is outside my area and I haven't read the Gligorijevic paper, but I am curious whether 86% accuracy should be considered good or good enough to be useful in this domain.  In some domains that might be useful.  In others it would be quite poor even if better that existing methods.

cgreene: @brettbj can you comment?

enricoferrero: @agitter My impression is that 86% accuracy is pretty good for this type of task given the complexity/heterogeneity of EHR data. It might be worth looking at Table 1 in the paper: they provide precision@K values for theirs and other methods and the performance improvement is often quite impressive. I think in information retrieval, precision@K might be a better performance metric than accuracy (but accuracy easier to report being a single number).

agitter: @enricoferrero thanks for clarifying.  I'm not sure that we need to expand on this in the paragraph, but I'm always curious how close to "solved" a problem is even after substantially improving on previous state of the art.

agapow: The issue of numbers occurs elsewhere in the paper (at least in the parts I wrote). I think that the number itself is not so much significant as is whether it improves on the prior state of the art and by how much. (And perhaps whether it "solves" the problem.) In a review paper for a general audience, I'd suggest that numbers might be superfluous except as illustrations.

agitter:
Merging #360 introduced conflicts here.

sections/03_categorize.md
brettbj: Not sure if I'm reading something incorrectly but:
For this they only look at ICD9 codes though (so wouldn't have as many issues with noise/ heterogeneity etc.) 

The task is:
Disease Phenotyping: Identifying diseases with similar contexts as the query disease.

But it looks like they define these as having shared GWAS hits (at 10^-5 significance)

Here's how it looks like they determine that:
Specifically, we selected 2,739 diseases found in the GWAS data and for each retrieved K nearest diseases, with k in {1, 2, 5, 10} . Each of the retrieved diseases was labeled as positive if it shares a gene with the query disease, and labeled negative otherwise, which is used as a proxy for having the same phenotype54. Then, we computed precision@K for each disease as a fraction of positive neighbors within the K retrieved ones, and report the average precision over all 2,739 diseases.

The selection of diseases is really important, for example I'd be really suspicious of any cancers etc. as they are likely to share genes? They also include things like 234 forms of Tuberculous and tuberculosis - both are caused by the same bacteria... is it novel to predict these are related? (that's almost 10%)

In any event it makes sense that predicting diseases that co-occur have similar genetic factors. I'm not sure why this method does so much better than the others tested but it is not surprising that a method is able to do well.

enricoferrero: @brettbj Thanks for reviewing the paper and feedback.

My understanding is that the authors try to address two tasks:

1.  Disease2Disease (D2D, or disease phenotyping), where the model is trained exclusively using raw EHR data and learn the disease representations using a method similar to word2vec where each record is a sentence and each disease is a word:
    > In this section we propose the disease2diseases (D2D) approach for learning disease representations, building upon ideas introduced by the recently proposed word2vec algorithm[...] the language model learns representations of diseases in a low-dimensional space using each patient dis- charge record as a “sentence” and the diseases within the record as “words”.

    To _validate_ their model they then use shared GWAS genes to asses how close two diseases really are (but the GWAS data is _not_ used during training): 
    > To train the D2D model, we used only EHR data (P dataset) [...] Specifically, we selected 2,739 diseases found in the GWAS data and for each retrieved K nearest diseases, with ∈K {1,2,5, 10}. Each of the retrieved diseases was labeled as positive if it shares a gene with the query disease, and labeled negative otherwise, which is used as a proxy for having the same phenotype54. Then, we computed precision@K for each disease as a fraction of positive neighbors within the K retrieved ones, and report the average precision over all 2,739 diseases...

2. DiseaseAndGenes2Diseases (DAG2D, or retrieval of disease - gene associations), a model using GWAS data to learn representations of diseases and genes:
    > The DAG2D model assumes that a subset of diseases from the training data D are associated with genes, where
the associations are provided by domain experts and considered as domain knowledge. We leverage this informa- tion by assigning a vector representation to each gene, and make use of disease contexts in the training data to jointly learn both disease vectors and gene vectors in the same low-dimensional space.

    This model is simply validated on a held-out test set:
    > To train the DAG2D model (for both of the tasks), we extended patient discharge records by associ- ating diseases to corresponding genes according to GWAS data [...] We randomly selected 20% of the diseases that have associated genetic data (the diseases found in the GWAS data set) and removed all their gene associations from the training data. Although we removed genetic information, these diseases are not removed from the original EHR data, so that we are able to learn their vector representations. Second, we learn DAG2D on the data set where the remaining 80% of diseases remained associated with their corresponding gene information data. The DAG2D model then contains low-dimensional representations of diseases and genes in the same embedding space, and we evaluated model performance by measuring how many of the removed genetic associations were correctly retrieved

So, given the performance measures and comparisons they provide I think the results of both models are quite impressive: in D2D they recapitulate genetics-based distances between diseases only using EHR data and in DAG2D they convincingly recover known disease genetic associations from data held out during training.

I'd be happy to try and improve my summary if it's unclear: any specific suggestion would be welcome/helpful.

@agapow Sure, I can remove the accuracy figure if that's uninformative. Do we have a paper-wide policy on this?

CC @cgreene @agitter

agapow: It's an overall style thing and my read of the journal guidelines. I'll leave it to @cgreene @agitter to make the call.

agitter: We don't currently have a paper-wide policy.  There are other sub-sections that include performance and accuracy values.

I'd still like to wait for a full review from @cgreene before we merge.  I suggest that we don't delay merging because of the accuracy statistic.  We will need to set a paper-wide and can apply it universally during editing.

@agapow can you please refer me to the relevant journal guidelines that you found?

agapow: http://rsif.royalsocietypublishing.org/content/information-authors-0

The key quote being:

> These articles should put such research in a wider context and be written in a style that will make them accessible to readers in a wide range of disciplines.

brettbj: @enricoferrero  - The issue I have is that I don't think finding these "diseases" have a common SNP (not even genome-wide significance) is interesting or impressive. What would a classifier that just looked for a common word get....? Or a classifier that just picked the closest ICD-9 code by number? The definition of different disease is flawed in my eyes. They looked for two ICD-9 codes with a common SNP, oftentimes these ICD-9 codes are a single disease presenting in a different part of the body. This is why phenotyping algorithms are usually used.

Check out the supplement: https://www.nature.com/article-assets/npg/srep/2016/160831/srep32404/extref/srep32404-s1.csv

 
image

or 

image

or

image

or 

image

or 

image

cgreene: I agree with @brettbj that the results - once considering the predictions from the supplement - are a bit uninspiring. The results support that this algorithm is better than some other algorithms for this specific task, but these do seem to be trivial predictions. An evaluation that focused on only items that couldn't be found by e.g. a shared word in the disease name might show that the method is still compelling. I think it's fine to mention the work, but I am a bit wary of trumpeting the results as the arrival of deep networks in this space. Let me try to make some revisions (we need to make a merge commit anyway) to add more nuance to our description, and let's see what people think.

cgreene:
@enricoferrero / @brettbj : see what you think of the aggregate changes now with my revisions. I tried to couch them a bit more in the difficult to eval sense (we may need to do this at other places too as we revise). The basic message is refined to: promising but needs more comparison to methods that could be used for these problems.

brettbj:
LGTM

sections/03_categorize.md
enricoferrero: @brettbj @cgreene thanks, that's fair enough - I admittedly didn't look in the supplementary material.

sections/03_categorize.md
```diff
@@ -143,21 +143,26 @@ This indicates a potential strength of deep methods. It may be possible to
...
+[@doi:10.1145/2939672.2939823]. Methods for neural embeddings learned from
+electronic health records have at least some ability to predict disease-disease
+associations [@doi:10.1038/srep32404]. However, the evaluation did not
+differentiate between simple predictions (i.e. the same disease in different
+sites of the body) and non-intutive ones. While promising, a lack of rigorous
+evaluations of the real-world utility of these kinds of features makes
+current contributions in this area difficult to evaluate. To examine the true
+utility, comparisons need to be performed against leading approaches (i.e.
+algorithms and data) as opposed to simply evaluating multiple algorithms on
+the same potentially limited dataset.
```
enricoferrero: @cgreene, I think your summary looks good - my only concern is that the second model (DAG2D) in #342 is not even mentioned anymore. To me, that is the most interesting part of the paper: being able to identify disease-associated genes from EHR data has big implications for drug discovery (and specifically for target identification). It might be that DAG2D suffers from the same flaws of D2D but my preference would have been to mention this particular application (which doesn't seem to be explored elsewhere).
CC @brettbj @agitter

sections/03_categorize.md
```diff
@@ -143,21 +143,26 @@ This indicates a potential strength of deep methods. It may be possible to
...
-have the natural document structure of clinical notes, and this issue was
-addressed by E. Choi et al. [@doi:10.1145/2939672.2939823], who built
-embeddings using a multi-layer network architecture which mimics the structure
-of claims data. While promising, difficulties in evaluating the quality of
-these kinds of features and variations in clinical coding practices remain as
-challenges to using them in practice.
+Several authors have created reusable feature sets for medical terminologies
+using natural language processing (NLP) and neural embedding models, as
+popularized by Word2vec [@tag:Word2Vec]. A goal of learning terminologies
+for different entities in the same vector spaec is to find relationships
```
enricoferrero: Typo, spaec --> space

cgreene:
@enricoferrero - feel free to file a new PR if you want to do more wordsmithing on it. Thanks!

PR #358, opened 2017-04-25
added sentence about comparing image vs tensor

bdo311:
@agitter Minor changes talking about next steps in variant calling, based on our conversation in #344

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -640,8 +640,11 @@ variant and fed these vectors into a fully connected deep neural network
 [@tag:Torracinta2016_deep_snp]. Unfortunately, this feature set required at
 least 15 iterations of software development to fine-tune, which will likely not
 be generalizable. Going forward, we foresee that variant calling will benefit
-most from optimized neural network architectures and better methods of encoding
-raw sequence and quality data (e.g. in tensor rather than RGBA format).
+more from optimizing neural network architectures than from developing features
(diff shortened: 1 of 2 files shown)


sections/04_study.md
```diff
@@ -640,8 +640,11 @@ variant and fed these vectors into a fully connected deep neural network
 [@tag:Torracinta2016_deep_snp]. Unfortunately, this feature set required at
 least 15 iterations of software development to fine-tune, which will likely not
 be generalizable. Going forward, we foresee that variant calling will benefit
-most from optimized neural network architectures and better methods of encoding
-raw sequence and quality data (e.g. in tensor rather than RGBA format).
+more from optimizing neural network architectures than from developing features
+by hand. An interesting and informative next step would be to test whether
```
agitter: It seems silly to put so much emphasis on a single sentence, but what do you think about this:

"An interesting and informative next step would be to test whether encoding raw sequence and quality data as an image, tensor, or some other mixed format produces the best variant calls, given identical network architectures and other parameters.  The advantages of encoding genomic data as images in order to initialize weights from networks pre-trained on generic, large-scale image datasets may diminish over time or prove to be a generally effective strategy."

The phrasing of my new sentence is awkward though.

bdo311: I agree with making the point about leveraging pretrained networks explicit. Moved some words around and added a reference to Xception

PR #359, opened 2017-04-26
Update README.md

brettbj:


dhimmel:
@brettbj the links are with respect to the `gh-pages` branch ([they'll work here](https://github.com/greenelab/deep-review/tree/2ba9353a599f08215aed91d9be87e5c1d22bc9b2)). They don't work if you're on `master`. Maybe the right thing to do is just to delete the hyperlink?

After this PR can you avoid making branches in this repo? Instead, fork and make branches in your personal repo. This is to keep this repository clean and to avoid confusion for most users who don't have access to `greenelab`.

brettbj:
I just did inline in Github for this example since it's just the readme but can do that in the future. 

I guess we can't add logic... but if not on master it should still go to the raw file right?

dhimmel:
The links you added don't go to the raw file. They go to the most recent version of the file as rendered by GitHub Pages.

brettbj:
But the other links should right?

dhimmel:
> But the other links should right?

What "other links" are you referencing?

There are three [relative links](https://github.com/blog/1395-relative-links-in-markup-files) in `output/README.md`: `index.html`, `github-pandoc.css`, `deep-review.pdf`. These links work as intended on the `gh-pages` branch, but do not work on `master` since these files are ignored. Any change should modify all three of these links. Thus far, I only see two options:

1. keep relative links as is
2. remove links altogether

The same issue exists in [`references/generated`](https://github.com/greenelab/deep-review/tree/master/references/generated) on `master`.

PR #363, opened 2017-04-28
Intro patches

agitter:
I corrected errors, added TODOs from #246, and fixed the references so that this will build without error.  I'd like to merge this clean version of #246 before one of us goes through and merges the two version of the intro.

PR #364, opened 2017-04-28
Drop undrafted sub-sections

### LINKED ISSUE #352: Dropping subsections
April 24, our stated date for dropping undrafted sub-sections, has arrived.  After scanning the open pull requests and remaining undrafted sub-sections in the outline, I propose we drop the following:

- Categorize
    - Has deep learning already induced a strategic inflection point for one or more aspects?
    - Will deep learning induce a strategic inflection point for categorization?
    - What unique potential does deep learning bring to this?
    - Where would you point your deep learning efforts if you had the time?
- Study
    - RNA-binding proteins
    - Signaling
- Treat
    - Modeling Metabolism and Chemical Reactivity

We don't have a miRNA binding draft in place, but @bdo311 has been writing excellent sections so I'd like to extend the deadline for that topic.

There are still several miscellaneous Discussion topics to cover as well as the abstract and conclusions.  Still, there has been amazing progress in the past two weeks!

agitter:
This version doesn't have my corrections from #363 so the Travis build will fail.  I'm not editing the RNA-binding protein section until we merge #356.

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -25,6 +25,9 @@ specific content to present.  Others may be important areas where we lack
 expertise, in which case we may acknowledge the application area but not
 dive into merits or weaknesses of individual methods.*
 
+`TODO: dropped signaling subsection, decide whether to discuss in intro
+and why it has received less attention, @tag:Chen2015_trans_species as example`
+
 ### Gene expression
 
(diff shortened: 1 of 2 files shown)


agitter:
There were no suggested changes to these minor updates so I'm merging.

Issue #109, opened 2016-10-14
Evaluating models.


cgreene:
One topic that's come up quite a bit (e.g. #13) is how do we best eval these models? In some cases (e.g. TF binding), AUC is unlikely to match the use case that we really care about. While this is true for **all** bioinformatics problems, we should probably raise the issue since we get the opportunity - as part of a headline review - to provide perspective.

agitter:
Yes, this is important.  The review #27 brought up precision-recall versus ROC but only briefly.  We can also provide some intuition about why ROC is especially bad with skewed classes and maybe even refer to alternatives beyond area under a curve.

A related topic that has come up a few times is the choice of a gold standard for evaluating models, which can be hard in some biomedical domains.

akundaje:
There is a nice review here in Nature Methods
http://www.nature.com/nmeth/journal/v13/n8/full/nmeth.3945.html that
clearly and intuitively showcases the problem with AUCs for unbalanced
learning

akundaje:
Yes confidence rated predictions are important and also in gold standard
datasets providing "ambiguous labels" or confidence on labels is useful and
a good practice. It helps with evaluation (you can weight the loss by the
confidence in the label) as well.

traversc:
@akundaje In the nmeth paper, their example of why ROC is bad in imbalanced datasets isn't exactly clear to me.  In fig 4a, they show an example of a case where there are only a few positive patients (e.g., 5 out of 100).  In their graph, several of those patients are predicted with zero false positives; and the next several patients were detected with ~10-20% FPR.  They state that this is bad without qualification, because 10-20% FPR means that there are more FPs than TPs.  

However, this performance would be considered really good in some cases, such as rapid screens.  For example, rapid HIV tests in clinical use have FPRs of ~40% or higher (http://www.bmj.com/content/335/7612/188).  

Therefore, I think they haven't proved their premise that AUC is always bad in imbalanced datasets.  I think it depends on the biological nature of the dataset.

gokceneraslan:
@traversc It's FDR that is not covered in ROC, not FPR. If one classifies 15 out of 100 people as positives, from which 5 are actually positive, then FPR=10/95=10.5% but FDR=10/15=66.7%.

agitter:
[...]

Issue #43, opened 2016-08-05
Convolutional neural network architectures for predicting DNA–protein binding


cgreene:
https://dx.doi.org/10.1093/bioinformatics/btw255

hussius:
This is a benchmarking paper focusing on convolutional network architectures for predicting transcription factor binding from sequence. They use the DeepBind (Alipanahi et al. 2015) architecture as a baseline, and proceed to vary some aspects of this architecture: the number of layers, the number of convolution kernels/filters, and the type of pooling.

One nice thing about this paper is that they provide a Docker image that they say is runnable on any GPU machine (although I did not try this). Everything is implemented in Caffe, which I have had some trouble installing on my Macs, so a Docker image is probably a good idea. 

They actually look at two different classification tasks (the quoted parts are verbatim from the paper): 

- the **motif discovery task** which "classifies sequences that are bound by a transcription factor from negative sequences that are dinucleotide shuffles of the positively bound sequences" and
- the **motif occupancy task** which "discriminates genomic motif instances that are bound
by a transcription factor (positive set) from motif instances that are not bound by the same transcription factor (negative set) in the same cell type"

Apart from the architectural parameters mentioned above, they seem to do parameter searches separately for each motif to try to optimize dropout rate, momentum, and the "delta" parameter in the AdaDelta optimizer. There are up to 690 classification tasks.

Some questions I had after reading this paper:

- Does it always have to be a binary classification task for each of the 690 motifs? Wouldn't it be quite natural to do multi-class classification? Too difficult / too little training data?
- It is shown in this paper as well as others that sequences are represented by "one-hot" matrices which are fed to convolutional layers. For images, the (2D) convolutional layers receive inputs of the form [image height, image width, # color channels]. For sequence classification it appears that the standard way is to also use 2D convolutions where height is set to 1 and the nucleotides are the "color channels". Is it self-evident that this is better than using a single color channel and treating the one-hot input matrix as a 2D "image"?
- The sampling of the parameter space is still quite sparse here, for understandable reasons (it takes time to run these models!).

akundaje:
[...]

PR #380, opened 2017-05-05
evaluation section writeup

jisraeli:
@agitter: Opening a PR for evaluation section writeup. Can you look over the draft below? I have yet to format the citations properly but would love to get your thoughts on the text.


agitter:
Thanks a lot, I'll look it over.  At first glance, some of the very TF-specific parts may need to be merged with the section @jacklanchantin is writing in #356.  Here, we can leave the parts that focus on the evaluation metrics, which will be general across domains.

You don't have to use the tags convention, only some sections do.  DOIs or arXiv ids formatted as in https://github.com/greenelab/deep-review/blob/master/CONTRIBUTING.md are perfectly fine.  There are also tags already for many of these papers, e.g. `[@tag:Zhou2015_deep_sea]` will work.

sections/06_discussion.md
```diff
@@ -6,22 +6,10 @@ can be further pruned.*
...
 
-*What are the challenges in evaluating deep learning models that are specific to
-this domain?  This can include a discussion of ROC versus precision-recall
-curves for the imbalanced classes often encountered in biomedical datasets.
-It could also mention alternative metrics that are used in specific sub-areas
-such as enrichment factors in virtual screening.  A lack of true gold standard
-data for some problems complicates both training and evaluation.  Confidence-
-weighted labels are valuable when available.*
-
-*Is progress in some biomedical areas slowed when new predictions (e.g. from
```
agitter: Do you plan to write anything about this part of the prompt?  If not, we should leave the italics so someone else can write something.  Or we collectively agree it isn't interesting and drop it.

sections/06_discussion.md
```diff
@@ -6,22 +6,10 @@ can be further pruned.*
 
 ### Evaluation
 
-*What are the challenges in evaluating deep learning models that are specific to
-this domain?  This can include a discussion of ROC versus precision-recall
-curves for the imbalanced classes often encountered in biomedical datasets.
-It could also mention alternative metrics that are used in specific sub-areas
-such as enrichment factors in virtual screening.  A lack of true gold standard
-data for some problems complicates both training and evaluation.  Confidence-
```
agitter: This part of the prompt came from @akundaje in #109:

> Yes confidence rated predictions are important and also in gold standard datasets providing "ambiguous labels" or confidence on labels is useful and a good practice. It helps with evaluation (you can weight the loss by the confidence in the label) as well.

I'd still like to include that point here.

sections/06_discussion.md
```diff
@@ -6,22 +6,10 @@ can be further pruned.*
...
...s (bp) and only a small fraction of them are implicated in specific biochemical activities. As a result, binary classification of genomic regions based on their biochemical activity results in small positive class and a large negative class with important consequences for model evaluation.  For example, based on processed ChIP-seq data, the number of reproducible transcription factor (TF) binding sites genome-wide is typically in the 1,000-50,000 range. ChIP-seq data allows TF binding site detection with ~200 bp resolution and thus less than 1% of the genome can be confidently labeled as bound for most transcription factors. Therefore, it is important evaluate the genome-wide recall and false discovery rates (FDR) of computational models of TF binding. TF binding site predictions may be used for targeted validation experiments and as features for models of downstream processes such as gene expression. Targeted experimental validation usually necessitates an FDR in the 5%-25% range. For downstream modeling, a low FDR for TF binding predictions is not as critical if the downstream models meet their respective recall and and FDR requirements; an FDR of 50% in this context may suffice.
```
agitter: We should reframe this to answer the question above: "What are the challenges in evaluating deep learning models that are specific to this domain?" That prompt arose from this discussion in #43 and #109 for context.  The TF binding problem is a very good motivating example, as you describe here, but we should present it as an example of the broader topic.  I suggest leading with a discussion of class imbalance and possibly other pitfalls that are sometimes ignored when evaluating predictive models.  Then you can transition to some of what you have here illustrating how it is a problem for TF binding.  That can include a few of these details about just how imbalanced this prediction task is.

It will also strengthen our case to refer to previous work that supports our claim that ROC is inappropriate for imbalanced datasets.  http://www.nature.com/nmeth/journal/v13/n8/full/nmeth.3945.html and http://dl.acm.org/citation.cfm?doid=1143844.1143874 are especially relevant.  http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0118432 came up before, but it isn't critical if we use those other two.

sections/06_discussion.md
```diff
@@ -6,22 +6,10 @@ can be further pruned.*
...
...w.nature.com/nmeth/journal/v12/n10/full/nmeth.3547.html#methods) and follow-ups such as [DanQ by Quang et al.](https://academic.oup.com/nar/article/44/11/e107/2468300/DanQ-a-hybrid-convolutional-and-recurrent-deep) report state of the art auROC and auPRC of classifying 17% of the genome with at least one TF binding site. [Qin et al.](http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005403) developed a deep learning model that outperforms both DeepBind and DeepSEA on their respective tasks and evaluations. However, a chromosome-wide evaluation of this model reveals that its recall at 50% FDR is less than 10% for most TFs. This suggests that recall at 10% FDR would be 0% or close to 0% for most TFs. These evaluations reveal that auROC and auPRC metrics are not good indicators of the performance needed for targeted validation. In fact, it is not unusual to obtain chromosome-wide auROC > 0.99 for TF binding predictions but a recall of 0% at 10% FDR. Thus, it is important to evaluate models of TF binding and other biochemical activities such as histone modifications and chromatin accessibility using a recall at fixed FDR thresholds on complete held out chromosomes.
```
agitter: The state of the art in TF binding can be merged into #356.  Our focus should be on evaluation metrics here, and possibly how the wrong metrics lead to incorrect conclusions about which method is the best for a task.

There are very good points in this paragraph to keep.  I think the most important is that very high auROC does not mean the problem is "solved" and can correspond to very poor scores from other, more relevant evaluation metrics.  The sentences like `In fact, it is not unusual to obtain chromosome-wide auROC > 0.99 for TF binding predictions but a recall of 0% at 10% FDR.` are especially relevant.

sections/06_discussion.md
agapow: Bit of a hand grenade, but given the whole ENCODE-Dan Graur fracas, man I so do not want to go near a statement like "only a small fraction of them are implicated in specific biochemical activities" and get tangled up into deliberations on what "biochemical activity" are.  Perhaps try: "only a small fraction can be unambiguously implicated  in specific biochemical functions".

akundaje: @agapow The phrase "biochemical function" is the one that is controversial. Biochemical activity is not. That is exactly what a ChIP-seq or most functional genomics assays measure i.e. biochemical activity which is not necessarily functional. I think you got the two mixed up. We definitely don't want to say anything about "function" or "functional". Just for reference, I was the primary author on that famous (infamous) ENCODE paper :).

agapow: Haha. Should have looked. 

However, I selected "function" deliberately in indicating that function was unknown. This changes the meaning of the given sentence a little, but in a direction that I thought was defensible. Happy to leave it as is.

aaronsheldon: As an external party reading the evaluation section I want to echo these comments. In the two proposed paragraphs I am having difficulty grasping the central question; is it:

1. The human genome is a particularly difficult problem domain for evaluation, because of the large imbalances in the data?
2. The ROC statistics can be misleading, or are perhaps inappropriate? 
3. That in general statistical theory and methodology, particularly around falsification of complex large parameter models, has not kept pace with the technological development in deep learning?

I think the last question is the most powerful of the three, and that the citations noted above speak to the concerns in the second and third question.

agitter: @aaronsheldon One of the main goals for this section was 2, with human genome annotation being one example of where ROC is inappropriate due to class imbalance.  Drug discovery is another we can include.

3 would be extremely interesting, but we are also growing quite short on time now that we are trying to submit in a week.  If you have thoughts on the topic, we'd be happy to include here.

aaronsheldon: @agitter Thank you, I'm not sure I can have results within the time constraints. I know John Myles White at Facebook has been vigorously debating the point for a couple years now, specifically questioning the applicability of Central Limit and LLN asymptotically driven traditional hypothesis testing to the validation of machine learning results. Getting a solid set of citations is going to take some pruning.

agitter:
@jisraeli we're going to submit in a couple days.  Will you have time to restructure this today or tomorrow?  The TF binding task is good for a specific example of evaluation issues, but we need to rework this to focus on metrics and more general evaluation considerations.

jisraeli:
@agitter: yes, I have time to work on this today. I will push a restructure by 6pm PT.

agitter:
@jisraeli Thanks!  You can check the original italicized prompts for some of the questions we wanted to address, in addition to my specific comments above.  The other discussion sections will also give examples of how we've been trying to describe things generally but use specific examples from the Categorize, Study, and Treat sections.

sections/06_discussion.md
jisraeli: I addressed this point in a subsection in (46dccc0)[https://github.com/jisraeli/deep-review/commit/46dccc05bface3e2ed661648d67f870b224b6056]

sections/06_discussion.md
jisraeli: I added it in italic in 46dccc0. I would suggest dropping it - I don't think I have much to say there beyond the obvious at this stage. Wonder what others think about this.

sections/06_discussion.md
jisraeli: I restructured the section in 46dccc0 to address 3 questions: 1) evaluation metrics, 2) formulation of labels, and 3) formulation of upper bounds

sections/06_discussion.md
jisraeli: I took out the discussion of TF binding results in 46dccc0 and tried to use it as a motivating example only

sections/06_discussion.md
```diff
@@ -6,9 +6,28 @@ can be further pruned.*
...
+What are the challenges in evaluating classification models that are specific to this domain? The human genome has 3E9 base pairs (bp) and only a small fraction of them are implicated in specific biochemical activities. As a result, classification of genomic regions based on their biochemical activity results in highly imbalanced classification. What are appropriate evaluation metrics that account for the label imbalance? The classification labels are formulated based on continuous value experimental signals - what is an appropriate procedure for formulating binary classification labels based on these signals? The experimental signals are only partially reproducible across experimental replicates - what is an appropriate upper bound for classification performance that accounts for the experimental reproducibility?
```
agitter: This intro works much better.  It might fit the rest of the discussion style better to turn some of these questions into statements.

The domain we cover is very broad - from medical records to genomics to protein structure - so I suggest we start even broader before zooming in to TFs.

What do you think of replacing the first sentence with something like:
> There are unique challenges to evaluating deep learning predictions in the biomedical domain.  We focus on TF binding prediction as a representative task to illustrate some of these issues.

We don't use "base pairs" again so you can drop the "(bp)" abbreviation.

After "highly imbalanced classification" please add "...classification, which also arises in other problems we review such as virtual screening for drug discovery".

sections/06_discussion.md
```diff
@@ -6,9 +6,28 @@ can be further pruned.*
...
+#### Evaluation metrics for imbalanced classification
+Small fractions of the genome are implicated in biochemical activities such as transcription factor (TF) binding and histone modifications. For example, less than 1% of the genome can be confidently labeled as bound for most transcription factors. Therefore, it is important to evaluate the genome-wide recall and false discovery rate (FDR) of classification models of biochemical activities. Targeted validation experiments of specific biochemical activities usually necessitate an FDR of 5%-25%. When predicted biochemical activities are used as features in other models, such as gene expression models, a low FDR may not be as critical if the downstream models can satisfy their evaluation criteria; an FDR of 50% in this context may suffice.
```
agitter: The TF acronym has been introduced before so you can use it here without reintroduction.

Do we need a reference for the 1% bound statement?  ENCODE or something else?

sections/06_discussion.md
```diff
@@ -6,9 +6,28 @@ can be further pruned.*
...
...as auPRC and auROC? auPRC evaluates the average precision, or equivalently, the average FDR across all recall thresholds. This metric provides an overall estimate of performance across all possible use cases which can be misleading for targeted validation experiments. For example, classification of TF binding sites can exhibit a recall of 0% at 10% FDR and auPRC greater than 0.6. In this case, the auPRC may be competitive but the predictions are prohibitive for targeted validation that can only examine a couple of predictions. auROC evaluates the average recall across all false positive rate (FPR) thresholds which is often a highly misleading metric in this domain. For example, consider a classification model with recall of 0% at FDR less than 25% and 100% recall at FDR greater than 25%. In the context of TF binding predictions where only 1% of genomic regions are bound by the TF, this is equivalent to a recall of 100% for FPR greater than 0.33%. In other words, the auROC would be 0.9967 and not indicative of the predictive utility for targeted validation. It is not unusual to obtain a chromosome-wide auROC greater than 0.99 for TF binding predictions but a recall of 0% at 10% FDR.
```
agitter: For auROC, how about: "auROC... highly misleading metric in class-imbalanced domains [@doi:10.1145/1143844.1143874 @doi:10.1038/nmeth.3945]" to add references to relevant literature.

Should we close this paragraph with a more general restatement of why this is a problem?  We could add a new sentence like:
> The classifier would appear to be nearly perfect based on auROC but has no practical utility to prioritize high-confidence binding sites for experimental validation.

sections/06_discussion.md
```diff
@@ -6,9 +6,28 @@ can be further pruned.*
...
+
+#### Formulation of classification labels
+Genome-wide continuous signals are commonly formulated into classification labels through signal peak detection. For example, peaks in ChIP-seq signal are used to identify locations of TF binding and histone modifications. Such procedures rely on thresholding criteria to define what constitutes a peak in the signal. This inevitably results in a set of signal peaks that are close to the threshold, not sufficient to constitute a positive label, and too similar to positively labeled examples to constitute a negative label. To avoid an arbitrary label for these example they may be labeled as “ambiguous”. Ambiguously labeled examples can then be ignored during model training and evaluation of recall and FDR. The correlation between model predictions on these examples and their signal values can be used to evaluate if the model correctly ranks these examples between positive and negative examples.
```
agitter: @jacklanchantin @bdo311 Do you think we should remove this from the TF binding section since similar discussion came up there?  We could refer to this section instead.

sections/06_discussion.md
```diff
@@ -6,9 +6,28 @@ can be further pruned.*
...
+
+#### Formulation of a performance upper bound
+Genome-wide signals across experiments can lead to different sets of positive examples. When experimental replicates do not completely agree, a 100% recall at 25% FDR is not possible. The upper bound on the recall is the fraction of positive examples that are in agreement across experiments. This fraction will vary depending on the available experimental data. For example, reproducibility for experimental replicates from the same lab is typically higher than experimental replicates across multiple labs. One way to handle the range of reproducibility is the use of multiple reproducibility criteria such as reproducibility across technical replicates, biological replicates from the same lab, and biological replicates from multiple labs.
+
+
+*Is progress in some biomedical areas slowed when new predictions (e.g. from
```
agitter: Upon further consideration, I agree with you about dropping the italics here.  I have nothing to add beyond the obvious.

agitter:
I noticed there is a merge conflict, but it is a trivial one.  I can take care of it myself once we finish this section.

sections/06_discussion.md
bdo311: The point about thresholding I think is succinct enough in the TF binding section. I can shorten the point about choosing negative training examples.

agitter:
@jisraeli We're planning to submit Wednesday.  Will you be able to make these small edits today?

Please also add yourself to the author list once you have read and approved the manuscript.  You can make a pull request on `author_contact_approve.md`.

jisraeli:
Yes I can push edits by 1pm PT. I will also read the rest of the manuscript today and add myself to the author list.

sections/06_discussion.md
jisraeli: Addressed TF acronyc in dfa475b. The 1% bound statement is based on internal analysis, there is a bit of nuance there in "confidently labeled". I don't know of a good reference that gets at the same nuance if we come up with one I am happy to cite it.

sections/06_discussion.md
jisraeli: Addressed these comments in dfa475b.

sections/06_discussion.md
jisraeli: I dropped the italics in fb65530.

sections/06_discussion.md
jisraeli: Yes, I agree with a more general restatement. I will commit this edit later today.

jisraeli: Addressed in 50d8234.

jisraeli:
@agitter: I believe I addressed all the comments and fixed the style checks and merge conflicts. Can you check if this is good to merge?

sections/06_discussion.md
agitter: We don't need a reference, I was curious if you happened to know of a good one.

agitter:
@jisraeli thanks for finishing this on time.  I'm merging now.

PR #389, opened 2017-05-08
Author approval

cgreene:
We will need to collect affiliations info for authors. @agitter - what do you think about asking contributors to fill in this form and note if they approve?


agitter:
@cgreene are we ready to merge this or should we also include the readme update to this pull request?

enricoferrero:
Questions/comments:
- How should we fill in the form? Each author creates a PR?
- Ideally all PRs will be merged before authors give approval.
- Re timelines: when do you plan to submit? We can relax the May 14th deadline based on that.

agitter:
Yes, each author will make a pull request to add a line to the table.  That would give us explicit confirmation of their approval.

We still intend to submit next week as far as I know.  We have pushed the deadline back several times already and the current deadline from the journal may be more firm than the previous deadlines.  Ideally we would merge all pull requests before seeking approval, but we may need to do things in parallel to keep that deadline.

PR #411, opened 2017-05-10
add brettbj to author list

brettbj:


agitter:
I like the enthusiastic approval @brettbj

PR #434, opened 2017-05-12
two typos

yfpeng:
Probably it is not worthy to open a pull request...


agitter:
Thanks @yfpeng, these typo-correcting pull requests are actually very helpful as we get ready to submit.

PR #437, opened 2017-05-13
shortened splicing section

bdo311:
Shortened and polished the splicing text.

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -138,7 +138,8 @@ required extensive human fine-tuning. Moreover, as in gene expression network
 analysis, interrogating the hidden nodes within neural networks will likely
 yield new biological insights into splicing. For instance, tissue-specific
 splicing mechanisms can be inferred by training networks on splicing data from
-different tissues, then searching for common versus distinctive nodes
+different tissues, then searching for common versus distinctive nodes, a
+technique employed by Qin et al. for tissue-specific TF binding predictions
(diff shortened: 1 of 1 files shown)


sections/04_study.md
```diff
@@ -134,57 +129,50 @@ exon skipping [@tag:Barash2010_splicing_code @tag:Xiong2011_bayesian]. With the
...
-networks for each tissue type and looking for common versus distinctive nodes
+multiple kinds of epigenomic measurements, as well as tissue identity and RNA
+binding partners of splicing factors. Deep learning is critical in furthering
+these kinds of integrative studies where different data types and inputs
+interact in unpredictable (often nonlinear) ways to create higher-order “features”,
+compared to earlier approaches that often assumed independence of features or
+required extensive human fine-tuning. Moreover, as in gene expression network
+analysis, interrogating the hidden nodes within neural networks will likely
+yield new biological insights into splicing. For instance, tissue-specific
+splicing mechanisms can be inferred by training networks on splicing data from
```
agitter: `Qin2017_onehot` studies TF binding, not splicing, right?  If so, should this state "mechanisms could be inferred"?  I think the idea is that one could use a model like Qin 2017 for tissue-specific splicing models.  We may even want to point out that Qin 2017 was for TF binding.

bdo311: Yeah it was for TF binding -- will update.

Issue #345, opened 2017-04-24
abbreviations: DNN, CNN, RBM, DBN, etc


alxndrkalinin:
Couldn't find any guidelines on abbreviations - are we going to introduce common ones like CNN in the beginning of the manuscript and use further on? Is there journal policy on this?

agitter:
Good question.  We haven't decided anything yet.  Defining the most common neural network abbreviations like those in the issue title in the introduction makes sense.  Others can be defined at first use.

@cgreene let us know if you have thoughts.

cgreene:
NDC -> National Drug Codes (came up in #339)

ICD -> International Classification of Diseases

I figure we'll need to go through and define acronyms the first time they appear. If you encounter some, feel free to note them here.

enricoferrero:
NLP: natural language processing (#342).
Plan to introduce it in Categorize/EHR, I can see it's also used in Study/Metagenomics.

agitter:
RF -> random forest

May decide to always write it out in full, but we should be consistent throughout.

agapow:
What I've spotted that appear to occur more than once:

EHR: electronic health record
RNN: recurrent NN
TF: transcription factor
MRI / fMRI: Magnetic resonance imaging
DBM: deep boltzman machine
DBN: deep belief network
MTL: multi-task learning

agitter:
AUC: area under the curve
ROC: receiver operating characteristic

Needed for #438 among others

agitter:
I'm going through these in #498 and adding 👍 once I cover them.

agitter:
Instead of abbreviating random forest as RF, which I don't care for, I'm using square brackets to write it out in a quote.  This will show up as a warning during builds, but we can ignore it.

PR #439, opened 2017-05-13
Shortened and polished TF section

bdo311:
I substantially shortened and revised the TF section to reflect things people have written in other sections. Tagging @jacklanchantin who wrote the initial draft.

I also addressed the TODO about RBPs by adding a sentence saying that deep learning methods applied to TFs will be broadly applicable to RBPs. I think this is sufficient but would love to hear people's opinions.

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -241,8 +241,8 @@ site), false positives or false negatives can result depending on the threshold
 chosen by the algorithm. Second, most methods predict binding of TFs at sites in
 isolation, whereas in reality multiple TFs may compete for binding at a single
 site or act synergistically to co-occupy it. Fortunately, multi-task models are
-rapidly improving at simultaneous prediction after a single training step
-[@tag:Zhou2015_deep_sea]. Third, it is unclear exactly how to define a
+rapidly improving at simultaneous prediction of many TFs' binding at any given
+site [@tag:Zhou2015_deep_sea]. Third, it is unclear exactly how to define a
 non-binding or "negative" site in the training data, since the number of
 positive binding sites of a particular TF is relatively small with respect to
 the total number of base-pairs in DNA (see Discussion).
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -188,139 +188,91 @@ disease, which will lead to novel diagnostics and therapeutics.
 
 ### Transcription factors and RNA-binding proteins
 
-Transcription Factor and RNA-binding proteins are key components for gene
-regulation, making them very important to understand for higher level
-biological processes. While high-throughput sequencing techniques such as
-chromatin immunoprecipitation and massively parallel DNA
-sequencing (ChIP-seq) have been able to accurately identify binding regions
-for DNA and RNA proteins, these experiments are both time consuming and
-expensive. In addition, the sequencing methods do not provide any sort of
-analysis on the proteins which would lead to a better understanding of
-the underlying process. Thus, there is a need to computationally predict
-and understand these binding regions de novo from sequences.
+Transcription factors (TFs) and RNA-binding proteins are key components in gene
+regulation and higher-level biological processes. While high-throughput
+sequencing techniques such as chromatin immunoprecipitation and massively
+parallel DNA sequencing (ChIP-seq) have been able to accurately identify targets
+for DNA- and RNA-binding proteins, these experiments are both time consuming and
+expensive. Thus, there is a need to computationally predict binding sites and
+understand binding specifities de novo from sequence data. In this section we
+focus on TFs, with the understanding that deep learning methods applied to TFs
+will also be broadly applicable to RNA-binding proteins.
 
-#### Transcription factors
+TFs are regulatory proteins that bind to certain genomic loci and control the
+rate of mRNA production. ChIP-seq and related technologies are able to identify
+highly likely binding sites for a certain TF, and databases such as ENCODE
+[@tag:Consortium2004_encode] have made freely available ChIP-seq data for
+hundreds of different TFs across many laboratories. In order to computationally
+predict transcription factor binding sites (TFBSs) on a DNA sequence,
+researchers initially used consensus sequences and position weight matrices to
+match against a test sequence [@tag:Stormo2000_dna]. Simple neural network
+classifiers were then proposed to differentiate positive and negative binding
+sites but did not show significant improvements over the weight matrix matching
+methods [@tag:Horton1992_assessment]. Later, SVM techniques outperformed the
+generative methods by using k-mer features [@tag:Ghandi2014_enhanced
+@tag:Setty2015_seqgl], but string kernel based SVM systems are limited by
+expensive computational cost proportional to the number of training and testing
+sequences. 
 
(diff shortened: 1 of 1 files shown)


jacklanchantin:
@bdo311 Thanks for the revisions Brian. I'm not opposed to making it shorter. I think your revisions are good.

sections/04_study.md
```diff
@@ -188,139 +188,91 @@ disease, which will lead to novel diagnostics and therapeutics.
...
+to account for such differences. For example, motifs may appear in either strand
+of a DNA sequence, resulting in two different forms of the motif (forward and
+reverse complement) due to complementary base pairing. To handle this issue,
+Shrikumar et al. [@tag:Shrikumar2017_reversecomplement] created a convolutional
+model that can find motifs in both directions. Since deep learning for protein
+binding prediction is still in early stages, we expect to see an increase in
+domain-specific architectures for this task.
+
+Despite these advances, several challenges remain. First, because the input data
+(ChIP-seq measurements) are continous and the output (whether or not there is TF
```
agitter: Something may have been lost in this contraction.  Can we make it clear that the output is the output of the deep learning system?  Making this binary is a modeling choice, not a necessity (though we have discussed previously why alternatives like regression aren't ideal either).

sections/04_study.md
```diff
@@ -188,139 +188,91 @@ disease, which will lead to novel diagnostics and therapeutics.
...
-noisy, resulting in a target class that may be incorrect dependent on the
-defined threshold.  Converting the task into a binary classification isn't
-completely accurate since the model may give a high probability of a binding
-site, but the ChIP-seq binding signal may be just barely over the binary
-classification threshold, resulting in potential false positive (or the signal
-may be just below the threshold, resulting in potential false negatives).
-
-Second, most datasets for TFBS prediction are separated by TF, requiring a
-separate model for each TF (i.e. binary classification on each TF sub-dataset).
-In reality, there may be multiple TFs binding at the same location and TF
```
agitter: The line starting with "In reality" was important to illustrate why this is a limitation.

sections/04_study.md
```diff
@@ -188,139 +188,91 @@ disease, which will lead to novel diagnostics and therapeutics.
...
+for TFBS prediction only assay one TF at a time, but in reality multiple TFs may
+bind at the same location in a cooperative or antagonistic manner. Deep learning
+can be useful in integrating different datasets and output one prediction
+[@tag:Zhou2015_deep_sea]. Third, it is unclear exactly how to define a
+non-binding or "negative" sites in the training data. Since the number of
+positive binding sites of a particular TF is relatively small with respect to
+the total number of base-pairs in DNA, we must choose a small subset of the
+total non-binding sites, resulting in some sort of bias over all of the actual
+negative sites. Regardless of the negative site selection, most datasets evenly
+balance the positive and negative binding sites and report auROC for the metric.
```
agitter: We've used AUC ROC elsewhere instead of auROC.  I'll note that in #345 and we can make ti consistent later.

sections/04_study.md
```diff
@@ -188,139 +188,91 @@ disease, which will lead to novel diagnostics and therapeutics.
...
+target TF. A more general domain transfer model across cell types would be more
+useful.
+
+Critically, deep learning can also provide useful biological insights into TF
+binding. Lanchantin et al. [@tag:Lanchantin2016_motif] and Shrikumar et al.
+[@tag:Shrikumar2016_blackbox] developed tools to visualize TF motifs learned
+from TFBS classification tasks. Alipanahi et al. [@tag:Alipanahi2015_predicting]
+also introduced mutation maps, where they could easily mutate, add, or delete
+base pairs in a sequence and see how the model changed its prediction. This
+would be very time consuming in a lab setting, but easy to simulate using their
```
agitter: "was easy"

sections/04_study.md
bdo311: I tried to capture this idea by saying that "in reality multiple TFs may bind at the same location in a cooperative or antagonistic manner". I'll add the part about multi task learning since I do think that's important. what do you think?

sections/04_study.md
bdo311: good point, I'll add that in.

sections/04_study.md
bdo311: I edited this to "Second, most methods
predict binding of TFs at sites in isolation, whereas in reality multiple TFs
may compete for binding at a single site or act synergistically to co-occupy it.
Fortunately, multi-task models are rapidly improving at simultaneous prediction
after a single training step [@tag:Zhou2015_deep_sea]."

Not sure if this says what we want it to say..

agapow:
Seems like a good call. Looks good.

sections/04_study.md
agitter: Your new phrasing handles this well

sections/04_study.md
agitter: Yes, I like this.  What is the part about a single training step referring to?  Do they initialize and then only train for a single epoch?  I didn't remember that in the paper, but it has been a while.

agitter:
I had one final question about the DeepSEA reference, then this looks ready to merge.

sections/04_study.md
bdo311: Oh no by single training step I meant that you train a single model to output predictions for multiple TFs at once, rather than training n models for n TFs. What would be a better wording for this?

agitter: We could say something like "simultaneous prediction of many TFs' binding sites [@tag:Zhou2015_deep_sea]."  We have the multi-task discussion section to elaborate on how this is done by training a single model.

jacklanchantin: @agitter @bdo311 I agree with this. The only issue I have is that while deepSEA is the only deep learning paper I know of that uses all TFs at once, TF binding prediction is only an intermediate step of their model and they are also predicting other epigenetic factors at the same time. So I'm not really sure it belongs in this section, but I wouldn't be opposed if it was.

agitter: @jacklanchantin By intermediate step, do you mean because they then stack another classifier on top to predict variant effects?  Their CNN predicts individual ChIP-seq tracks for specific TFs (in a specific cell type), right?

If you have a better phrasing, we can use it instead of what I suggested.

jacklanchantin: @agitter Yes, I meant because they stack the final classifier on top. They do predict the individual ChIP-seq tracks for specific TFs in a specific cell type, but they report the TFBS, DHS, and HM AUC scores compared to previous techniques, so I think that it's ok to put it here.

I think I talked too much about how most models use TFs separately, which is true, but the data for all TFs at once is there from ENCODE (which is what they use in DeepSEA), so I think it's good to just mention DeepSEA here and show that they did TF binding prediction the right way.

bdo311: Yep I agree with just mentioning it briefly. How does "simultaneous prediction of many TFs' binding at any given site [@tag:Zhou2015_deep_sea]" sound?

agitter:
Everything looks good to me.  Thanks again for revising so much of the Study section!

PR #436, opened 2017-05-13
Shortened and polished gene expression section

bdo311:
My general feeling is that the sections within "Study" all have a lot of overlapping content (since they were written by different contributors) so I'm trying to remove redundancy.

PR #442, opened 2017-05-14
added batch effects reference

bdo311:
From #436

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -671,8 +671,7 @@ networks trained only on CpG or sequence information. Similar deep learning
 methods have been applied to impute low-resolution ChIP-seq signal from bulk
 tissue with great success, and they could easily be adapted to single cell data
 [@tag:Qin2017_onehot @tag:Koh2016_denoising]. Deep learning has also been useful
-for dealing with batch effects, which can severely affect single-cell data
-[@tag:Shaham2016_batch_effects].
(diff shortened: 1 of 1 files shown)


sections/04_study.md
```diff
@@ -670,7 +670,9 @@ several percentage points of improvement compared to random forests or deep
 networks trained only on CpG or sequence information. Similar deep learning
 methods have been applied to impute low-resolution ChIP-seq signal from bulk
 tissue with great success, and they could easily be adapted to single cell data
-[@tag:Qin2017_onehot @tag:Koh2016_denoising].
+[@tag:Qin2017_onehot @tag:Koh2016_denoising]. Deep learning has also been useful
+for dealing with batch effects, which can severely affect single-cell data
```
agapow: Well, batch effects can severely affect lots of areas, so the "single cell" qualification may not be needed.

cgreene: Reference for dealing with batch effects? Might be nice to guide the reader.

cgreene: Ahh - after another glance - I see what you've done here. I was interpreting the reference as related to the ", which" clause.

cgreene: I agree with @agapow now - I think it's clearer without that clause.

agitter:
@bdo311 Can you please remove the TODO at the end of the single cell section? That was a reminder to add this paper.

PR #452, opened 2017-05-16
Remove enhancer TODOs and other revisions

agitter:
@agapow I removed the TODOs here and made a few other changes, most notably shortening the last paragraph.

As I re-read this, I was wondering about the in silico mutagenesis connection I asked you to add earlier in regards to Basset (Kelly et al.)
> This study also featured a useful interpretability approach (analogous to *in silico* mutagenesis [@doi:10.1093/nar/gkm238]) introducing known protein binding motifs into sequences and measuring the change in predicted accessibility.

I originally thought your description of their interpretation strategy sounded like ISM.  Now I'm not so sure.  Should we remove the parenthetical?

### DIFF
--- a/sections/04_study.md
+++ b/sections/04_study.md
@@ -299,40 +299,37 @@ papers`
 
 ### Promoters, enhancers, and related epigenomic tasks
 
-`TODO: There's inevitably a lot of overlap been this and the transcription
-binding section. Maybe in the long term think about restructuring to one big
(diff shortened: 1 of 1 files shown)


agapow:
It's no great loss if it goes. So, do it.

agitter:
I'll cut the ISM phrase and merge.  Thanks for reviewing.

PR #455, opened 2017-05-16
Add Jack Lanchantin and Yanjun Qi

jacklanchantin:
Dr. Qi and I worked on the TFBS section together.


cgreene:
Thanks! Did a quick search of the issues/discussion and can confirm that both contributed to GitHub discussion. @jacklanchantin confirms @qiyanjun's contribution to writing of the section.

jacklanchantin:
@cgreene Thanks! I added our names to the end, but I wanted to let you know that we are fine with whatever ordering you and @agitter feel is right

cgreene:
Yea - the order in that file doesn't relate to author order. I think there is some within-category randomization being discussed. Thanks!

PR #456, opened 2017-05-16
Authorship approval of dhimmel

dhimmel:


dhimmel:
@cgreene can you let me know if there is any funding I should acknowledge.

cgreene:
@dhimmel GBMF 4552 from the Gordon and Betty Moore Foundation should cover you. If there is any overlap with the work that you've done on cognoma (i.e. re-using systems that you initially configured for Cognoma for this) you could add NIH NS095411

PR #457, opened 2017-05-16
A minor change to make the URL work

j3xugit:
I added one space after "url:http://www.predictioncenter.org/casp12/rrc_avrg_results.cgi" to make this link work.


agitter:
@dhimmel We should leave this as-is I think.  Do you agree?

@j3xugit The citation will be converted to a numeric link with the URL shown in the references.  You can search for "CASP12" in https://greenelab.github.io/deep-review to see how it will appear.  The URL in the references works.

dhimmel:
> to make this link work

@j3xugit work where? It's already working.

PR #458, opened 2017-05-16
Training on simulated data

agitter:
I added a discussion section on simulated data.  I thought about adding some thoughts on how GANs fit in, but this is long enough and we're out of time.

I believe this was the last major TODO that required new text as opposed to editing or references.


agitter:
Thanks for reviewing.  I need to re-read DeepBound (doi:10.1101/125229) once more before merging.  They may have trained on purely simulated data and evaluated on real data, which I'd like to highlight.

agitter:
I added the additional DeepBound line so I'm merging now.

PR #368, opened 2017-05-01
Interpretability section

AvantiShri:
I have finished fleshing out the interpretability section.  Anshul has looked over it and given his approval. Sorry about the delay; I had a flare up of RSI and  had to use speech recognition software

PR #469, opened 2017-05-17
Addressed outstanding TODOs in the interpretability section

AvantiShri:
@agitter thanks a lot for addressing the issues with my first pull request (https://github.com/greenelab/deep-review/pull/368). Here is a second pull request with the remaining TODOs addressed.

Regarding approval for the manuscript, I focused my attention on "Transcription factors and RNA-binding proteins" and "Promoters, enhancers, and related epigenomic tasks" under 04_study, as I have the most familiarity with those. I have a few comments (tagging @akundaje and @jisraeli for their thoughts):

The reverse complement paper was cited as "Shrikumar et al" - since Peyton and I were co-first on this paper, can this please be cited as "Shrikumar & Greenside et al."?

Can the reference to @tag:Shrikumar2016_blackbox please be changed to the more up-to-date @tag:Shrikumar2017_learning?

I feel like DeepSEA should be mentioned under "Promoters, enhancers, and related epigenomic tasks" because I'm reasonably sure they included epigenomic datasets from ENCODE in their prediction tasks. It would be most natural to mention it alongside basset, i.e. somewhere near this line: "Most broadly, Kelley et al. [@doi:10.1101/gr.200535.115] trained CNNs on DNA accessibility datasets". I have also included the interpretability approach mentioned in the context of Basset under perturbation-based approaches in the interpretation section (note that DeepSEA also did in-silico mutagenesis, which is yet another perturbation-based approach to interpretability)

"This work also included a promising and simple method for interpretability (randomly substituting bases in a recognized promoter region, then checking for a change in recognition) that would be useful to exploit more widely." - I added this method under "perturbation" approaches in the intepretation section. Maybe remove the "useful to exploity more widely" comment, as it is after all a variant of a method that is indeed exploited more widely?

"Singh et al. [@doi:10.1101/085241] also used batch normalization" - wait, Basset (Kelley et al.) also uses batch normalization. The use of batch normalization isn't particularly unique - why mention it specifically?

Thanks so much for being so patient with me, and sorry for the delay!


agitter:
@AvantiShri Thanks for addressing these TODOs and reviewing the Study section text.  In general, some of the confusing issues you note (e.g. DeepSEA and Basset placement) is due to the TF and enhancers sections being written by different people at different times.  We wanted to refactor and merge these but ran out of time.

Some of these changes are very minor and can be made now.  I'll do it myself.  For some of these changes, I suggest that we create an issue and fix them for version 2 of the manuscript.  @cgreene and I have already started final editing for submission.  I do want to respect your very valid concerns and submit something you are comfortable with though.

@cgreene how extensively have you proofread the Study section already?  I don't want to make changes if you have already polished and finalized parts of it.  Is it safe for me to merge the interpretation text?

references/tags.tsv
```diff
@@ -198,6 +198,7 @@ Seide2014_parallel	doi:10.1109/ICASSP.2014.6853593
 Setty2015_seqgl	doi:10.1371/journal.pcbi.1004271
 Selvaraju2016_grad	arxiv:1610.02391
 Shaham2016_batch_effects	arxiv:1610.04181
+Shapely_wikipedia	url:https://en.wikipedia.org/wiki/Shapley_value
```
agitter: Is the Hazewinkel reference on the Wikipedia page suitable instead of the Wikipedia page itself?

sections/06_discussion.md
```diff
@@ -115,19 +115,30 @@ divided into perturbation-based approaches and backpropagation-based approaches.
 
 ##### Perturbation-based approaches
 
-These approaches make perturbations to individual inputs and observes the impact
-on the output of the network. Zhou and Troyanskaya [@tag:Zhou2015_deep_sea]
-scored genomic sequences by introducing virtual mutations at each position and
-quantifying the change in the output. Ribeiro et al. [@tag:Ribeiro2016_lime]
+These approaches make perturbations to parts of the input and observe the impact
+on the output of the network. Zhou & Troyanskaya and Alipanahi et al.
```
agitter: Move references to match the placement (I'll change directly).

sections/06_discussion.md
```diff
@@ -326,15 +339,39 @@ that produce similar results to the whole input when passed through an encoder.
...
-`TODO: Are there any brief final thoughts you would like to add? This paper is
-part review, part perspective piece so we have the opportunity to speculate
-about the future or push certain for certain research directions. For example:
-Is the criticism of deep learning as a black box, uninterpretable approach
-exaggerated given all of these methods? Are there certain types of features or
-domains (in genomics, healthcare, etc.) where these methods still fall short?
-Are there practical steps that would improve adoption of appropriate model
-interpretation? Maybe closer integration into the most popular deep learning
-frameworks, as applicable?`
+#### Final thoughts
```
agitter: What do you think about calling this "Future outlook"?

sections/06_discussion.md
agitter: Done

references/tags.tsv
AvantiShri: Yes, looks like it!

sections/06_discussion.md
AvantiShri: Sounds good to me

AvantiShri:
Thanks @agitter, I have replied to your questions. I don't plan to make any more commits (you can feel free to change my "yes (pending minor feedback)" in the author approval section to a "yes", now that I know you have seen the feedback.

references/tags.tsv
agitter: Turns out the Wikipedia link was to a massive math book.  I'll use reference [1] here instead https://www.encyclopediaofmath.org/index.php/Shapley_value

cgreene:
@agitter : feel free to merge. I have a dead tree version - not electronic yet - and will cross-check changes with the current version.

references/tags.tsv
agitter: @dhimmel I can't find a suitable URL for this.  A DOI seems unlikely for a reference from the 50s.  I'm using `url:http://oai.dtic.mil/oai/oai?verb=getRecord&metadataPrefix=html&identifier=AD0604084` for now and will keep hunting.

dhimmel: See https://doi.org/10.1515/9781400881970-018 (DOIs can be retroactively issued). I'll update the Wikipedia reference with the DOI.

dhimmel: [Added](https://en.wikipedia.org/w/index.php?title=Shapley_value&type=revision&diff=780889801&oldid=770615222) 3 DOI references to the Wikipedia page. I'm impressed all these old book chapters have DOIs!

agitter: Wow!  That's great.

PR #470, opened 2017-05-17
Remove speech recognition TODO

agitter:
I'm removing this TODO because I have not found a better reference for the conversational speech recognition error trajectory.  https://developer.ibm.com/watson/blog/2016/04/28/recent-advances-in-conversational-speech-recognition-2/ is a reasonable article but not much better than what I have.  https://github.com/syhw/wer_are_we has good references but may not be stable.  I was not able to source the Eric Horvitz tweet I considered earlier.


agitter:
Cross reference #369 for a time stamp

Issue #473, opened 2017-05-17
TF and enhancer comments


agitter:
These comments are carried over from #469, which contains more details:
- [x] The reverse complement paper was cited as "Shrikumar et al"
- [x] Change @tag:Shrikumar2016_blackbox to @tag:Shrikumar2017_learning
- [x]  DeepSEA should be mentioned under "Promoters, enhancers, and related epigenomic tasks" 
- [x] Remove the "useful to exploit more widely" comment
- [x] Clean up phrase "Singh et al. [@doi:10.1101/085241] also used batch normalization"

Tagging @AvantiShri, I'll work on these.

agitter:
#475 closed this issue

PR #474, opened 2017-05-18
Remove duplicate references

agitter:
I found cases where we had a preprint and published version or a later preprint for the same work.  I removed `doi:10.1101/039800`, `arxiv:1602.05629`, and `Shrikumar2016_blackbox` (which addresses part of #473).


agitter:
I verified these were resolved correctly in the latest manuscript build on the master branch.

PR #475, opened 2017-05-18
TF binding and enhancer revisions

agitter:
Addresses the issues in #473.  This intentionally builds on the edits in #474, one of which was needed for #473 as well.

@AvantiShri I tried to fix the problems you found while making the smallest edits possible given how close we are to submission.  We plan to keep making revisions while the paper is under review, and it would be great to have your help in restructuring the TF and enhancer sub-sections.


cgreene:
LGTM 👍

PR #476, opened 2017-05-18
Author table standardization

agitter:
@dhimmel I'm testing out a notebook to parse the author contact table.  I thought it might be easiest to programmatically generate a funding statement, affiliation map, etc.  This doesn't necessarily need to be merged to the master branch, but I wanted to see if you had any thoughts.

Currently my email address handling is a hack and Unicode characters aren't converted correctly.


dhimmel:
@agitter I'd highly recommend just converting `author_contact_approve.md` to `author_contact_approve.tsv`. This will obsolete all the problematic parts of notebook (md table to pandas) as well as the unicode errors IIUC.

agitter:
@dhimmel Converting to tsv via pandoc or with my own code? I haven't worked with pandoc or markdown tables much.

dhimmel:
> Converting to tsv via pandoc or with my own code? I haven't worked with pandoc or markdown tables much.

You can probably just copy the [GitHub rendered table](https://github.com/agitter/deep-review/blob/f19b1060be5f49a922b209042c924e2f7d398f8f/author_contact_approve.md) into a text file. It'll be a semi-manual step, but you'll only have to do it once.

I'm happy to do it if you run into any issues.

agitter:
Okay.  I was trying to avoid any manual steps, but maybe I'm making the problem needlessly difficult.

agitter:
This is still ugly, but functional.  I now have the author table in pandas for us to create affiliation mappings and so forth.

dhimmel:
@agitter curious why you're pursuing hacks rather than just converting the author table to TSV?

Current method relies on `pypandoc`, which isn't in our environment and is a problematic dependency (in my experience). Is there any reason you want to stick with markdown for the author table source?

agitter:
@dhimmel My assumption was that we should leave the primary data source as `.md`, which I believe we selected originally because we thought it would be easier for contributors to edit in GitHub.  We had many delimiter-related errors with the `tags.tsv` file.  Now that most of the authors have been added, we could convert to `.tsv` permanently.  The latest commit drops the `pypandoc` dependency, but is still not a robust converter.

My main goal with this is to have a way to automatically update parts of the manuscript, like affiliation mapping and funding acknowledgements, from our source author table.  I expect our author list will change during revision so I wanted to avoid doing that manually.

If there's a better way to do this, we can close this pull request.

dhimmel:
> My main goal with this is to have a way to automatically update parts of the manuscript, like affiliation mapping and funding acknowledgements, from our source author table. I expect our author list will change during revision so I wanted to avoid doing that manually.

What you need is a machine readable and human editable table format. I think TSV provides the right balance between machine readability and human editability. Ultimately, if you find a reliable way to convert markdown tables to dataframes, then it's no big deal. But this section:

```python
mod_file = orig_file.replace('|---------------|-----------|--------|-----------------|-------|---------|----------------|\n','')
mod_file = mod_file.replace('  ',' ')
mod_file = mod_file.replace(' | ','\t')
mod_file = mod_file.replace('| ','')
mod_file = mod_file.replace(' |','')
mod_file = mod_file.replace('|','')
# temporary until I look into character encoding problems
mod_file = mod_file.replace('Ã¤','ä')
mod_file = mod_file.replace('Ã¼','ü')
```

makes me nervous. Seems prone to breaking.

>  We had many delimiter-related errors with the tags.tsv file.

Some contributors had difficulty with spaces versus tabs. If you're really worried, you could even use a `.csv` extension and pipe-delimit. But I also think it's reasonable for contributors to be able to manually edit a TSV.

Also not sure what's causing the encoding issues. If you point me to where it goes wrong, I can take a look.

agitter:
@dhimmel Thanks for the suggestions.  I'm going to drop this for now and focus on the text.  I think switching to TSV will be the best solution.  Some of the entries have commas so I prefer that over CSV.

agitter:
In lieu of the suggesting of moving to TSV, I'm reducing the scope of the pull request.  Now it only contains minor fixes to the author table: removing hidden special characters and other standardization.

PR #478, opened 2017-05-18
Remove TODO in Treat Section

blengerich:
Added two references to requested paragraph.

PR #477, opened 2017-05-18
Section: Continuous collaborative manuscript drafting

dhimmel:
@agitter and @cgreene sorry for getting this in last minute.

I used [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/) (one sentence per line), since I think they're better than word wrap. I'm sure codeclimate will have a fit.

PR #479, opened 2017-05-18
Count references by type

dhimmel:
Adds to the build output something that looks like:

```
References by type:
doi      274
arxiv    106
url       40
pmid       4
Name: standard_citation, dtype: int64
```


dhimmel:
Use these stats to fill in `XX` in https://github.com/greenelab/deep-review/pull/477

agitter:
Cool!  We'll have to watch these counts if we have any new pull requests.  #478 looks like it only uses existing references.

PR #480, opened 2017-05-18
Proofread abstract and introduction

agitter:
I am ignoring abbreviation standardization during this initial pass.  I am separately compiling a list of references that will require manual formatting updates.


agitter:
@cgreene Per your suggestion I didn't reflow to minimize the number of lines affected.  If you prefer, I can easily reflow.

PR #484, opened 2017-05-18
Study proofreading

agitter:
This isn't ready to merge, but I have to step away for a while.  I wanted to share which parts I've edited.

I would like to add the reference for the statement "accuracy of Xiong et al.’s deep learning-based framework" but didn't have time to look up the correct Xiong et al. paper.

PR #485, opened 2017-05-19
Enable templating with jinja2

dhimmel:
This PR enables templating, so that including `{{reference_counts.doi}}` will insert the number of DOI references. See https://github.com/greenelab/deep-review/pull/484#issuecomment-302744491

### DIFF
--- a/build/environment.yml
+++ b/build/environment.yml
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - bioconda::wkhtmltopdf=0.12.3
   - conda-forge::pandoc=1.19.2
+  - jinja2=2.9.6
   - pandas=0.19.2
   - python=3.6.0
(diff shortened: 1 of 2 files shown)


agitter:
@dhimmel LGTM.  I knew you would go with [2](https://github.com/greenelab/deep-review/pull/484#issuecomment-302744491) 😄 

Now that templating is enabled, we *could* use it to generate things like a funding statement.  Something like "We acknowledge NIH grants `{{nih.statement}}`" where `nih.statement` is "AI116794 (B.K.B), DP2GM123485 (A.K.), ...".  That may not be the easiest or best approach.

dhimmel:
> Now that templating is enabled, we could use it to generate things like a funding statement.

Good point! I'm all in favor of automating whatever we reliably can.

PR #489, opened 2017-05-19
Proofread Discussion

agitter:
Not finished yet, but I want to check how this version builds after adding a reference.


agitter:
This is ready for review now.  There are multiple sections that feel too long and repeat points made earlier in the paper.  I did not undertake a major restructuring, but we may want to revisit it during revisions.

PR #492, opened 2017-05-21
Proofread Categorize

agitter:
I removed the link to the CVPR17 data.  The [paper](https://arxiv.org/abs/1705.02315) provides that Google URL and points to https://www.cc.nih.gov/drd/summers.html for more details, but there aren't instructions at Dr. Summers's NIH site.  @yfpeng, it would be great to add this resource back in if you can explain the data access to use.

@alxndrkalinin needs to help us resolve the DOI for `@tag:Abramoff2016_dr`, which is the same as `@tag:Pratt2016_dr`.  It is probably `doi:10.1167/iovs.16-19964` but please confirm.  I removed the TODO for @traversc because he hasn't been available lately, and we have a new section in discrimination.

During revision, we should think about how to reduce the redundancy between this section and the Discussion.  We make some of the same points about data limitations and transfer learning in both places.

### DIFF
--- a/references/tags.tsv
+++ b/references/tags.tsv
@@ -1,6 +1,6 @@
 tag	citation
 Abe	doi:10.1101/gr.634603
-Abramoff2016_dr	doi:10.1016/j.procs.2016.07.014
+Abramoff2016_dr	doi:10.1167/iovs.16-19964
 Agarwal2015_targetscan	doi:10.7554/eLife.05005
 Alipanahi2015_predicting	doi:10.1038/nbt.3300
 AltaeTran2016_one_shot	doi:10.1021/acscentsci.6b00367
--- a/sections/03_categorize.md
+++ b/sections/03_categorize.md
@@ -2,7 +2,7 @@
 
 In a healthcare setting, individuals are diagnosed with a disease or condition
 based on symptoms, the results of certain diagnostic tests, or other factors.
-Once diagnosed with a disease an individual might be assigned a stage based on
+Once diagnosed with a disease, an individual might be assigned a stage based on
 another set of human-defined rules. While these rules are refined over time, the
 process is evolutionary rather than revolutionary.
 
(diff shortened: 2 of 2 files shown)


alxndrkalinin:
@agitter sorry for the delay with response, you're right – for `@tag:Abramoff2016_dr` it should be `doi:10.1167/iovs.16-19964`

agitter:
@alxndrkalinin Thanks for confirming!

agitter:
@cgreene I'm merging now.  While re-reading this, I recalled conversations with colleagues who were cautious about deep learning the clinic due to errors and inaccuracies in the EHR.  Do you think this is a valid concern?  If so, do we need to note it more explicitly (but briefly) in the "challenges" sub-section here?

I'm not going to made any more edits until I have confirmation that you are done submitting pull requests.  Feel free to edit any section.

cgreene:
Ok - I have chatted with @laurakwiley who is an expert in EHRs + phenotyping. She may contribute this today or tomorrow. If she does, I've asked her to please also add her name to the authors table.

laurakwiley:
Ok so the more I look at this section the more I want to add.  I think there is a lot more to say on the data quality issue and how it affects deep learning (like 1-3 paragraphs, but spread over multiple sections). However with traveling and such I just cant make it happen by tomorrow's deadline. If you can wait until Friday, I could add it, or you could decide it isn't a priority. For what it's worth the major gaps are: non-random missingness due to fragmented care, and clinical presentation (briefly touched on); data quality measures; laboratory test differences (both reference ranges across labs and over time), and health policy HIPAA/Common Rule etc legal issues. Let me know how you would like to proceed.

cgreene:
@laurakwiley : can you submit a partial PR? We could add you as a co-author then, get the order nailed down, and then merge your final set of changes towards the end of the week.

PR #500, opened 2017-05-23
Added updates and citations for EHR biases

laurakwiley:
This is about 70% of what I could include. I'm somewhat inclined to leave this as is so y'all can submit today. The major piece of the literature not included is the work around EHR data quality.


laurakwiley:
Sorry newbie to this. fixed things and will close and send new request.

Issue #381, opened 2017-05-05
URLs in references


agitter:
We've be using URL references for two purposes:
1) Blog posts, online news articles, etc.
2) Conference papers

In both cases, do we need to add some additional information to the references?  Case 2 is the greater concern.  I'm wondering whether we need to start working on a local BibTeX file to resolve these.

dhimmel:
@agitter I'm on vacation until May 16. It's unlikely I'll be able to get to this until then. But here are the options:

1. Leave as is. The journal can fill in the URL references at a later time.
2. Look into automatic methods for URL citation metadata extraction. There is something called [Greycite](http://greycite.knowledgeblog.org/) that may be able to do this. However, it's uptime and reliability is questionable.
3. Manually add metadata (either as CSL JSON or bibtex) for the URLs. This may be good if the articles are in Google Scholar (although those bibtex records often have errors).

Option 3 will give best results. If you want to add more metadata for URLs, I'd begin option 3 in my absence.

agitter:
Thanks @dhimmel.  I like the idea of 2 but expect it will slow us down.  We may go with 1 so that we don't delay the submission and can explore 3 if time permits.

I'll leave the issue open to remind me to work on 3 if I can.

agitter:
Cross-referencing #387, a book with a URL citation.

dhimmel:
Greycite is working at this time. [This API call](http://greycite.knowledgeblog.org/json?uri=http://papers.nips.cc/paper/5954-convolutional-networks-on-graphs-for-learning-molecular-fingerprints) for `@url:http://papers.nips.cc/paper/5954-convolutional-networks-on-graphs-for-learning-molecular-fingerprints` returns:

```json
{
   "URL":"http://papers.nips.cc/paper/5954-convolutional-networks-on-graphs-for-learning-molecular-fingerprints",
   "title":"Convolutional Networks on Graphs for Learning Molecular Fingerprints",
   "issued":{
      "date-parts":[
         [
            2015
         ]
      ]
   },
   "author":[
      {
         "family":"Duvenaud",
         "given":"David K."
      },
      {
         "family":"Maclaurin",
         "given":"Dougal"
      },
      {
         "family":"Iparraguirre",
         "given":"Jorge"
      },
      {
         "family":"Bombarell",
         "given":"Rafael"
      },
      {
         "family":"Hirzel",
         "given":"Timothy"
      },
      {
         "family":"Aspuru-Guzik",
         "given":"Alan"
[...]

PR #504, opened 2017-05-23
Replace URL citations with ids

agitter:
Addresses three papers from #381


agitter:
The more we can handle automatically the better!  I'll merge and confirm they looked okay in the built manuscript.

agitter:
These showed up reasonably well in the built manuscript.  PMID 27330264 isn't perfect, but it's no worse than others.

dhimmel:
Looks like we should also add hyperlinks or IDs to pubmed, in addition to arxiv refs.

PR #505, opened 2017-05-23
Manually override citation information

agitter:
@dhimmel I followed your instructions in #381, but please let me know if there are errors.  I'd like to confirm this format is correct before moving on.  I didn't know which values are supported by the CSL for the `type` field.

The last set of manual updates will copy and modify the existing JSON entries.  I'll append those to this file.

I'm not sure whether which branch to merge this with.  I'll close and refine if master is preferable.

### DIFF
--- a/manual-citations.json
+++ b/manual-citations.json
@@ -1,0 +1,427 @@
+[
+ {
+  "citation_id": "url:http://www.bioinf.at/publications/2014/NIPS2014a.pdf",
+  "type": "article-journal",
+  "title": "Deep learning as an opportunity in virtual screening",
+  "container-title": "Neural Information Processing Systems 2014: Deep Learning and Representation Learning Workshop",
+  "URL": "http://www.dlworkshop.org/23.pdf?attredirects=0",
+  "author": [
+   {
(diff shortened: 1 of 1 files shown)


dhimmel:
Looks good. I haven't actually validated it, but I don't see any issues.

> I didn't know which values are supported by the CSL for the type field.

Here are the [valid options](https://github.com/citation-style-language/schema/blob/4846e02f0a775a8272819204379a4f8d7f45c16c/csl-data.json#L12-L46).

> I'm not sure whether which branch to merge this with. I'll close and refine if master is preferable.

We'll want `manual-citations.json` (let's rename to `manual-references.json`) to be in the `references` directory of the `master` branch, i.e. https://github.com/greenelab/deep-review/tree/master/references. Remember that the `references` branch is entirely autogenerated.

agitter:
Thanks for inspecting this.  I'll close and open a new pull request in master.

I'm only using
```
"type": "article-journal"
"type": "paper-conference"
"type": "report"
```
so that will be okay.

dhimmel:
Moved to #506

PR #511, opened 2017-05-24
Update author_contact_approve.md

traversc:
Hi all, I was away on a trip for the past week or so and just came back.  Apologies for the late response, but I would like to be re-added to the author list if it is still possible.  Thank you.


agitter:
@traversc Just in time!  Please also follow [these instructions](https://github.com/greenelab/deep-review/issues/369#issuecomment-303191214) **tonight** to create an account with the journal as well.

traversc:
@agitter Thank you!  I created an account with journal, per those instructions.

agitter:
Thanks. Please leave the account name in that other thread if you haven't
already.

PR #514, opened 2017-05-25
Fix additional references

### LINKED ISSUE #381: URLs in references

agitter:
We can close #381 once this merges.  I'll have to check the final output to confirm everything worked as intended.

### DIFF
--- a/references/manual-references.json
+++ b/references/manual-references.json
@@ -773,7 +773,7 @@
       {
         "given": "Aaron",
         "family": "Courville"
-      },
+      }
     ],
     "type": "webpage",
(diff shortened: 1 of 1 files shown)


agitter:
@dhimmel It looks like I was able to fix my JSON formatting errors.  I'll wait for you to look at the build before merging.

agitter:
Merging

PR #520, opened 2017-05-25
Add competing interests.

cgreene:
Instructions from the journal say: 

`If a competing interest is declared please provide full details of this in a competing interests section before the acknowledgements section of your main manuscript file.`

See https://github.com/greenelab/deep-review/issues/518


agitter:
@cgreene That reminded me to change our author counter.  @dhimmel and I will automate that during revision.

I'll review this after we hear back from everyone in #518.

PR #521, opened 2017-05-26
Update README to note new manuscript status

cgreene:
@agitter : I wanted to get this in before you go out of touch for a bit. If you think this is good, let me know. I will wait to merge until we get the bioRxiv information. At that point, I'll include that in the readme as well and merge.


agitter:
@cgreene I added the bioRxiv link.  I think we're ready to merge and tag.

PR #528, opened 2017-06-05
note availability of preprint

cgreene:


cgreene:
Good point! Otherwise it will become strangely self-referential :)

Issue #523, opened 2017-05-27
This point on domain adaptation doesn't make sense to me


jisraeli:
```Without cell-specific features, another solution could be to use domain adaptation methods where the model trains on a source cell type and uses unsupervised feature extraction methods to predict on a target cell type``` in the `Transcription factors and RNA-binding proteins` subsection doesn't make sense to me. Non cell-specific features, absent cell-specific information, have the same distribution in source and target celltypes. Domain adaptation is used to normalize for differences between the source and target distribution, so I'd argue the opposite - that cell-specific features with different source and target distributions must be used to take advantage of this technique. Am I misunderstanding this point?

agitter:
@jisraeli I'm traveling and will look into the intentions of this sentence later this week.

agitter:
@jacklanchantin I'm looking back at our discussion in #356 to recall the context of the quoted sentence above.  I believe this sentence was written in regards to the [comment](https://github.com/greenelab/deep-review/pull/356#issuecomment-299262230)
> You can briefly mention the additional complication that domain adaptation methods will also be imperative to account for the fact that the same TF can have very different cofactors across conditions/cell types. Hence, sequence features that optimal in one condition may result in very poor generalization in a new condition. This area of research will be challenging but exciting.

Is that correct?  Let's clarify this domain adaptation sentence.

jacklanchantin:
@agitter sorry, I must've missed this. I think there are 2 main aspects of the domain adaptation area (corresponding to 2 different problems)

1. In the case of cross species, since the sequences are different, we can use domain adaptation methods such as the methods used to predict clothing reviews based on a training set of jewelry reviews. Similar task, but the context is different. In this case, with regards to Anshul's comment, hopefully the unsupervised methods will be able to pick up the fundamental differences between domains which are useful for prediction.

2. In the case off cross cell lines, the sequences are almost identical, so we need more advance techniques of domain adaptation (I don't think anyone really knows exactly what yet), or using cell-type specific inputs such as histone modifications in addition to the sequences.

So, I think what is written is correct, but it doesn't exactly address Anshul's comment. Should I add a clarification of that?

akundaje:
I think the best way to clarify this is to use a well-defined example (like
the one you mentioned above) so that the statement is understood in the
appropriate context.

- Anshul.

jacklanchantin:
@akundaje do you think we should mention both cross cell line and cross species? Or just cross cell line?

akundaje:
[...]

PR #532, opened 2017-06-07
Update domain adaptation in TFBS section

jacklanchantin:


sections/04_study.md
```diff
@@ -220,13 +220,19 @@ prediction at the sequence level, they often cannot predict binding patterns for
...
-a target cell type. TFImpute [@tag:Qin2017_onehot] predicts binding in new cell
-type-TF pairs, but the cell types must be in the training set for other TFs.
-This is a step in the right direction, but a more general domain transfer model
-across cell types would be more useful.
+methylation, or gene expression. Without cell-specific features, we need models
+which can handle the differences between the domains. For example, if we have
+the ChIP-seq data for a particular TF from a mouse, but we want to predict the
+binding locations for that same TF on a human, we need to transfer the knowledge
+of training on the mouse to the human. This has been done in many other areas
+such as sentiment analysis [@doi:10.1.1.231.3442]. For example, we can train on
```
agitter: This isn't a valid DOI.  Can you please change the citation to `[@url:http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.231.3442]`?

jacklanchantin:
@agitter thanks for pointing that out. So they just note that each domain has labelled data for testing purposes, but it's only used to train one domain. As noted in section 3.3, they first train an autoencoder on the source domain and a separate autoencoder on the target domain. They then train an SVM classifier on top of the autoencoder in the source domain (where they use the labelled examples), and use that same SVM classifier to test on the target domain. So in the target domain, they aren't using any labelled samples, just an autoencoder to reconstruct the samples. The idea is that hopefully each autoencoder will learn the useful high level features for that class and the SVM will be able differentiate regardless of class. 

Does that make sense?

There are many more works in this area which have improved this model, but I don't think we have to note them here.

agitter:
@jacklanchantin Thanks for clearing that up, I hadn't read the paper in full.  I'll review this text more carefully now so we can merge

@jisraeli let us know if this addresses your questions in #523.

jacklanchantin:
I can write more on this subject and related work if needed. I'm not sure how short you want it though.

agitter:
@jacklanchantin I think we can keep it fairly short.  My goal would be just a few sentences to introduce the idea, the new citation, and how the cited method (or the general domain adaptation concept) applies in binding.  You don't need to write more, and if anything I might suggest ways to contract a couple sentences.

jacklanchantin:
@agitter ok sounds good to me. Thanks

agitter:
@jacklanchantin Thinking more about the overall message we want to send about this type of domain adaptation, would you agree that idea from the Glorot 2011 paper could be relevant but it would be very hard to accomplish in practice for TF binding?  In the human and mouse example, we do have sequence differences (as opposed to different human cell types).  However, I haven't seen any unsupervised method that I would trust to extract the relevant information about binding sites given only whole genome sequence as input.  If you agree, we may want to present this by saying the the Glorot 2011 paper gives an example of how domain adaptation can work, but it is much more challenging to construct the autoencodor (or equivalent) in DNA sequence domain than the text reviews domain.

jacklanchantin:
@agitter I definitely agree with the fact that the mouse/human adaptation is probably a much harder problem than those used in Glorot 2011. But to be honest, I don't know how difficult it would be by using only sequence information. I would hope that some experimentation with autoencoders may be able to shed some light on it, but I don't really have the biological background to say how different the sequences across species would be in this case. Based on my knowledge, it seems like it would be very difficult because the sequences will be almost the same, but I don't know for sure. 

I think your approach is good though. Introducing the work done in other fields, but saying that this exact approach may not be useful for a biomedical sequence task.

agitter:
Sounds good.  Now that I understand what we want to say, I'll look more carefully at the phrasing.

jisraeli:
@jacklanchantin @agitter: Thank you for taking a look at this, I am just now catching up on this PR.

I would like to suggest the following approach after `While deep learning-based models can automatically extract features for TFBS prediction at the sequence level, they often cannot predict binding patterns for cell types or conditions that have not been previously studied. `:

* Two considerations in the context of predicting in unseen celltypes are the use of celltype-specific features to enable generalization and the use of domain adaptation techniques to reduce overfitting to celltype-specific features.
* In the case of predicting cross species, domain adaptation may be used to reduce overfitting to differences in distribution of genomic sequence features between the species. For example, one may train a model on human genome sequence and regularize it via domain adaptation using mouse sequence. However, since only a small fraction of the genome differs between human and mouse, it is not clear how well domain adaptation would work in this case.

What do you think?

jacklanchantin:
@jisraeli @agitter I think that's a good idea, but I don't know if I did a good enough job introducing the differences between cross cell type and cross species. I only briefly introduced cross species. Do you think this is enough to summarize the differences/challenges within those 2 different cases? @agitter , I'm not sure how much you wanted to add to this section.

agitter:
I agree with the direction @jisraeli proposed.  We'll need to place the TFImpute discussion, perhaps before these points about unseen cell types or conditions.

@jacklanchantin Regarding the differences between cross cell type and cross species, we can briefly include a phrase to make it explicit that in cross cell type there is no sequence difference, only (epi)genomic features that differ.

jacklanchantin:
@agitter @jisraeli so should we mention domain adaptation for cross cell type?

agitter:
@jacklanchantin I think to do that we need multi-modal data in one of the cell types, right?  I like the domain adaptation reference you added because conceivably the strategy could be applied to sequence alone.  Sequence-only and sequence + complementary data (in the style of the DREAM challenge) approaches can both be discussed.

jacklanchantin:
In other words, I'm not entirely sure how different cross cell type and cross species sequences are. I supposed its highly dependent on the certain TF, cell type, and species, but for the most part they are probably the same?

akundaje:
If you for example focused on sequence underlying DNase peaks (which is one
of the multimodal inputs) across 2 very divergent cell types, you would
notice that the sequence has very different distributions and this
auxiliary mode (DNase) can help you adapt using sequence without using any
labels (TF peaks) in the test cell type.

-Anshul.

jacklanchantin:
@akundaje thanks! I figured that would be the case, but I don't want to make assumptions since I haven't read anything about it. Do you know of papers on this? I'm planning on doing some basic distribution analysis on GM12878 vs K562

akundaje:
No there are no papers on this. Nobody has really used the data in this
way. We are working on methods leveraging such information. I would
recommend you check HepG2 vs GM12878. The core pioneer factors that are
prevalent in accessible sites are very different in these cell types and we
see strong evidence of this issue in the DREAM challenge.

-Anshul.

jacklanchantin:
@akundaje ok great thanks a lot

agitter:
@jacklanchantin We should wait to merge this until we can add something about FactorNet (#549).

@akundaje Does the FactorNet preprint imply the DREAM preprint is coming soon?  Once that is available, it will influence how we can discuss cross cell type prediction.

agitter:
@jacklanchantin and @akundaje let's revive this conversation so we can see if this is ready to merge.  There have been some new papers since the initial pull request that affect how we want to edit this section.

akundaje:
Sure. I can make some edits tomorrow.

Anshul.

jacklanchantin:
@agitter @akundaje - Sure, I can edit this section. I'm pretty familiar with the FactorNet paper. Anything else to add? I'm not sure if you want me to write the domain adaptation part. I could add what @jisraeli wrote above. I'm pretty familiar with the idea of transferring across cell types. We had a paper (Transfer String Kernel for Cross-Context DNA-Protein Binding Prediction) which did this, but it didn't use any deep learning methods. But I could still write about deep learning models which could work in this scenario. 

@akundaje  - let me know if you want me to write a summary if the DREAM challenge if you don't have time. Obviously you're more familiar with it though.

akundaje:
Jack - Go ahead and do an editing pass. I can do a pass thereafter.

Anshul.

sections/04_study.md
```diff
@@ -216,17 +216,41 @@ positive binding sites of a particular TF is relatively small with respect to
...
+different. TFImpute [@tag:Qin2017_onehot] predicts binding in new cell type-TF
+pairs, but the cell types must be in the training set for other TFs. This
+is a step in the right direction, but a more general domain transfer model
+across cell types would be more useful. The main issue with predicting
+cross cell types using domain adaptation on sequence data is that
+the sequences in both cell types are likely identical. As a result,
+it would be necessary to also incorporate cell type specific data.
+
+Similar to cross cell type prediction, we also have TF binding data for
+a certain species (e.g mouse), but not one of interest
```
agitter: Period missing: `e.g. mouse`

agitter:
@dhimmel this pull request is from before the manubot update and there are likely extensive conflicts in `04_study.md`.  Would you suggest we follow your approach from #724 and replay the commits here in a new pull request?

@jacklanchantin and @akundaje I think it will be easiest if we merge this pull request first and then make a new one starting from the current draft.  I'd like to refine this section a little more but approved so that we can merge and continue working on it.

dhimmel:
> I think it will be easiest if we merge this pull request first and then make a new one starting from the current draft

I'd love to reflow the document to be one sentence per line before we open any new PRs. It seems like most active PRs will be merged soon. Does this make sense? Let's push to merge things, leaving follow up work till after the reflow?

> Would you suggest we follow your approach from #724 and replay the commits here in a new pull request?

You don't need to open a new PR. You can merge / rebase and resolve the conflicts in this PR. If you want to open a new PR that's fine, but it won't make the conflict resolution any easier... unless perhaps you should manually replay your changes if that's easier then fixing the git conflicts?

jacklanchantin:
@agitter @dhimmel So can we merge what we have now and then I'll make the new edits?

agitter:
> I'd love to reflow the document to be one sentence per line before we open any new PRs. It seems like most active PRs will be merged soon. Does this make sense? Let's push to merge things, leaving follow up work till after the reflow?

That's fine with me.  I'll create issues as needed so that we don't forget about any of the lingering revisions in between pull requests.

> unless perhaps you should manually replay your changes if that's easier then fixing the git conflicts?

@dhimmel that's what I was debating.  We can try to fix the git conflicts first before resorting to that.

> So can we merge what we have now and then I'll make the new edits?

@jacklanchantin Yes, this is ready to merge (leaving further revisions for after the end-to-end reflow), but you'll need to resolve the conflicts with the current version of `sections/04_study.md` first.  Are you comfortable attempting the conflict resolution yourself?  I'll formally review and merge after that.  (Also, see my comment about a missing period above.)

jacklanchantin:
@agitter is it ok to merge now?

jacklanchantin:
Great. Thanks a lot Anthony!

PR #534, opened 2017-06-07
Add anchor links to headers in the HTML version

rgieseke:
This pull request adds links to section headers making it easier to link to a specific part of the manuscript.

The MIT-licensed AnchorJS library is used:
<https://www.bryanbraun.com/anchorjs/>

Styling would be adjustable.

The pandoc author recommends a simple JS solution, so I thought this would be the best way (editing the Pandoc templates might be an alternative):
https://github.com/jgm/pandoc/issues/2723


dhimmel:
Confirming `build/assets/anchors.js` is [this file](https://github.com/bryanbraun/anchorjs/blob/064abdd0987f305933ec4982af6d0c1cf2fd0814/anchor.min.js)? Do we want to keep the filename the same for clarity?

I haven't tested what the final product looks like, but I think this will be a valuable addition! Thanks @rgieseke

rgieseke:
Correct, I embedded the `anchor.min.js` code plus the 2 lines to initialize and select the elements to change. As the JS is embedded in the final HTML I think the filename isn't so important.

I'll add a comment to make it clearer where the code came from.

The output looks like this on mouseover, other symbols are possible:

screen shot 2017-06-07 at 21 10 24

PR #517, opened 2017-05-25
Improve affiliation formatting

dhimmel:
https://github.com/greenelab/deep-review/issues/516#issuecomment-304100123

PR #541, opened 2017-06-11
Convert author table from md to tsv

agitter:
@dhimmel I'm picking up our discussion from #476 and #517 about fully automating the author-related information such as author counts, affiliations, and funding.  As a first step, this converts the author table to a `tsv` file so that it can be easily imported into pandas.

I would also like your thoughts on where to modify the build code to parse the author table and generate strings for the jinja2 templates.  Should I create a new `authors.py` file and keep `references.py` as the main script?


dhimmel:
@agitter, the GitHub diff says "Binary file not shown" for `author_contact_approve.tsv`. What's the encoding of this file? We'll want UTF-8.

dhimmel:
> Should I create a new authors.py file and keep references.py as the main script?

Perhaps we should create a function in `authors.py` and import that into `references.py` so we don't clutter up `references.py` too much more. I can do this task, unless you'd like to.

dhimmel:
Looking better. I'd remove `(s)` in `Affiliation(s)`. For best machine readability, I go with column names in all lowercase and underscore-separated words. Really whatever is fine, and you'll have to pick the desired balance between machine and human column names.

agitter:
@dhimmel The initial conversion preserved all of the content from the original `.md` file.  I'm happy to make more modifications toward machine readability now that we plan to automate this further.  I'll change the column headers, and let me know if you have other suggestions.

In terms of code design, I'm staring to think I may need an `Authors` class.  I envision instantiating once by loading the `.tsv` file into pandas.  Then I can add methods to:
- get the number of authors
- generate a funding statement
- generate the author list and number affiliations

I haven't thought much about the design though.  I'm very receptive to any ideas you have.

dhimmel:
Okay, if you create a PR to autogenerate the author info, I'll provide feedback by review. Also happy to help if you get stuck.

dhimmel:
> In terms of code design, I'm staring to think I may need an Authors class

I think creating a class may be overkill. You can probably just work with dataframes.

agitter:
> Okay, if you create a PR to autogenerate the author info, I'll provide feedback by review. Also happy to help if you get stuck.

Thanks.  I'll start with something simple and we can iterate.

PR #545, opened 2017-06-17
Update build notes

agitter:
Should I also add a line to the build readme stating that the environment does not currently support Windows?  In Windows, `conda env create --file environment.yml`  gives

```
NoPackagesFoundError: Package missing in current win-64 channels:
  - wkhtmltopdf 0.12.3*
```

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -12,3 +12,4 @@ conda env create --file environment.yml
 ```
 
 Activate with `source activate deepreview`.
+Currently, the environment requires Linux.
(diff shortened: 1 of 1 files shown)


dhimmel:
It [looks like](https://anaconda.org/bioconda/wkhtmltopdf/files) `wkhtmltopdf` is currently only built for linux. So the environment most likely will fail on Mac and Windows. We can look into fixing this if that's a priority. For now, maybe just add that the environment is linux only.

agitter:
Fixing it is not a priority.  I agree we should just add a note, which I did in 369b2a8657a300c1ea524b538ed80b0fcf09dfd9.

PR #546, opened 2017-06-17
Template for author counts

agitter:
This is an small pull request to confirm I am using the templates correctly.  It only computes the number of authors.  The rest of the author information remains hard-coded.  If this looks okay, I'll add more functions to `authors.py` to return funding information, affiliations, etc. in the `author_info` dict.

@dhimmel I assumed we would want to write the author information to `stats.json`.  Do you agree?


build/authors.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def get_author_info(author_file):
+    """
+    Load the author table and return a dict with author information such as
+    the number of authors.
+    """
+    author_df = pandas.read_table(author_file)
+
+    author_info = dict()
```
dhimmel: Let's use a `collections.OrderedDict` here since we want the `stats.json` to have a deterministic ordering.

build/references.py
```diff
@@ -150,6 +152,10 @@ def get_standard_citatation(citation, cache, override):
 print('References by type:')
 print(ref_counts)
 
+# Author table information
+root_dir = pathlib.Path('..')
+stats['authors'] = get_author_info(root_dir.joinpath('author_contact_approve.tsv'))
```
dhimmel: Can we rename `author_contact_approve.tsv` to `authors.tsv`?

dhimmel: Maybe a bit cleaner to do:

```
authors_path = pathlib.Path('../authors.tsv')
```

since we don't use root_dir anywhere else?

build/authors.py
```diff
@@ -0,0 +1,17 @@
...
+
+import pandas
+
+
+def get_author_info(author_file):
+    """
+    Load the author table and return a dict with author information such as
+    the number of authors.
+    """
+    author_df = pandas.read_table(author_file)
```
dhimmel: Cool that pandas is compatible with pathlib!

agitter:
Good suggestions, I made those changes.

PR #567, opened 2017-07-07
Add a subsection: Text applications in healthcare

luzh2:
We added a new subsection to summarize the deep learning approaches in bio text mining area. We also added a new figure. The figure is correctly displayed in the HTML file, but not in the PDF file.

PR #569, opened 2017-07-08
explicitly use the utf8 encoding to read and write files

yfpeng:
Fix the UnicodeEncodeError while building the source on Center OS.

### DIFF
--- a/build/citations.py
+++ b/build/citations.py
@@ -2,7 +2,6 @@ import collections
 from hashlib import blake2b
 import pathlib
 import re
-import sys
 
 import base62
 import bibtexparser
--- a/build/citations.py
+++ b/build/citations.py
@@ -2,6 +2,7 @@ import collections
 from hashlib import blake2b
 import pathlib
 import re
+import sys
 
 import base62
 import bibtexparser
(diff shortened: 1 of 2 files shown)


agitter:
Can you please remove the modifications to `tags.tsv` and `03_categorize.md` so we can review those separately in #567?  I saw that pull request but will need more time to look it over.

Did the build error occur within the `deepreview` conda environment or a different environment?

@dhimmel these changes to `references.py` and `citations.py` look okay to me.

yfpeng:
> Can you please remove the modifications to tags.tsv and 03_categorize.md so we can review those separately in #567? I saw that pull request but will need more time to look it over.

Fixed.

> Did the build error occur within the deepreview conda environment or a different environment?

It was within the deepreview conda.

build/citations.py
```diff
@@ -2,6 +2,7 @@
 from hashlib import blake2b
 import pathlib
 import re
+import sys
```
dhimmel: Unused import?

yfpeng: > Unused import?
Removed

> What is Center OS? http://www.centeros.com/ or https://www.centos.org/ ?
 Sorry. It is CentOS.

> Instead it defaults to whatever `locale.getpreferredencoding()` returns.

On the system we are using, it returns "ANSI_X3.4-1968"

Issue #565, opened 2017-07-06
Duchenne/Becker Muscular Dystrophy and splicing


SiminaB:
I think the discussion of Duchenne/Becker MD and splicing should be a little clearer and also include specific references. In particular, Duchenne is the more severe form and Becker the less severe form of the same kind of disorder (but in reality it can be more of a continuum - see last paragraph here), both of them being caused by mutations to the same gene (dystrophin) - the mutations in Becker are usually "in frame" versus the ones in Duchenne which usually cause frameshifts (or are nonsense mutations) and lead to no functional protein being made.

I'm not familiar with Duchenne and Becker being classified as "splicing-related diseases" per se. The splicing *I think* comes in more at the therapy development stage in this case. For example, one class of therapies being considered are "exon skipping therapies" (one recently got conditionally approved by the FDA) which skip the mutated exon to get back in frame. The idea is to try to convert the more severe Duchenne phenotype to be closer to a less severe Becker phenotype. Another class of therapies being considered that is related to splicing is to use gene-therapy with a smaller version of dystrophin ("micro-dystrophin"). At 79 exons, dystrophin is the largest known human gene, so it is difficult to deliver the entire gene, which is why researchers are working on delivering a smaller version that still has the most important domains.
Here is one review on exon skipping (there a bunch): http://www.sciencedirect.com/science/article/pii/S0169409X15001015 
And a review on genetic approaches, including mini-dystrophin:
https://www.nature.com/nrg/journal/v14/n6/full/nrg3460.html

It could be that there are splicing issues for some specific types of Duchenne/Becker as well, as the disease course can still be quite heterogeneous despite it being a single-gene disorder. While there is overall a strong correlation between the mutation disrupting the reading frame and the phenotype, there are cases which are "genetically Becker but clinically Duchenne" as well as vice-versa and intermediate phenotypes. However, this should be stated clearly, for instance, in the example with the skipping of exon 31 in Becker.

cgreene:
Pinging @agitter - do you recall who addressed that section? It'd be nice to bring them in here. Subsequent to that discussion (or before), @SiminaB if you'd like to edit the edit the text please file a pull request making the changes you'd like to see/agree on so that you can receive credit for the contribution.

agitter:
I believe @bdo311 wrote that section.  Is that correct @bdo311?

[...]

PR #566, opened 2017-07-07
Made some changes in Intro

SiminaB:
Wanted to have the introduction be a bit clearer about what deep learning is and how it falls into the midst of the "usual methods." Please check for accuracy.

PR #588, opened 2017-07-25
Changes to splicing subsection

SiminaB:
As we discussed in https://github.com/greenelab/deep-review/issues/565, I created a pull request to change the examples. I also added a line on therapies that try to modulate splicing.


SiminaB:
I used the version of the Intro file from the current greenelab/deep-review/master, but I just did copy/paste, so there may be some small changes. Sorry, I'm still figuring out branching.

sections/04_study.md
```diff
@@ -1,4 +1,4 @@
-## Deep learning to study the fundamental biological processes underlying human disease
+﻿## Deep learning to study the fundamental biological processes underlying human disease
```
agitter: Did something change in this line?  The text looks the same to me.

agitter:
@SiminaB thanks for making this pull request.  It should be fast to review and merge.  The only commit to focus on is 518c17a9a103c6905c4ecfc06b89d5bf253236d3, right?  I'm copying @bdo311 as well.

I think what happened with the intro is that you rolled back to before I edited in intro in #540 so those changes have been reapplied.  I should be able to clean that up and will focus on `04_study.md`.

The other commits with your intro changes showed up here because the commits in #566 were in your `master` branch.  When you created the `splicing-pull` branch, [it was created from your local master](https://github.com/SiminaB/deep-review/network), which was ahead of `greenelab:master`.  Making the branch like you did here should work in the future once your `master` is synced with `greenelab:master` again.  We'll do a squash merge so none of that will matter here.

sections/04_study.md
SiminaB: I don't see a change either. Maybe some spacing somewhere?

SiminaB:
@agitter Yes, just https://github.com/greenelab/deep-review/commit/518c17a9a103c6905c4ecfc06b89d5bf253236d3. Again, sorry for my poor branching! I think you're right about what happened.

sections/04_study.md
```diff
@@ -138,7 +141,7 @@ instructive in that clever sources of data, not just more descriptive models,
 are still critical.
 
 We already understand how mis-splicing of a single gene can cause diseases such
```
bdo311: Maybe "lead to" instead of "cause"? I think it softens up the causality a little bit -- also I'm definitely not an expert on limb girdle muscular dystrophy but it's probably safe to assume that there are some cases not caused by splicing.

bdo311:
I have one comment but everything else looks good to me -- thanks!

sections/04_study.md
```diff
@@ -97,13 +97,16 @@ Pre-mRNA transcripts can be spliced into different isoforms by retaining or
 skipping subsets of exons or including parts of introns, creating enormous
 spatiotemporal flexibility to generate multiple distinct proteins from a single
 gene. This remarkable complexity can lend itself to defects that underlie many
-diseases [@tag:Scotti2016_missplicing]. For instance, in Becker muscular
-dystrophy, a point mutation in dystrophin creates an exon splice silencer that
-induces skipping of exon 31. A recent study found that quantitative trait loci
+diseases. For instance, splicing mutations in the lamin A (LMNA) gene can lead to specific
```
agitter: We can put `*LMNA*` in italics

sections/04_study.md
agitter: @bdo311 The authors of the review use the language that splicing-related mutations can cause disease, so I'm comfortable with `cause` here.  If you still object, I don't feel strongly about it so we could make the change.

agitter:
@SiminaB This pull request is almost ready to merge.  Can you please see the comments above and make the minor changes if you agree?

SiminaB:
So the change ended up just being italicizing the gene name. To me "lead to" and "cause" are actually synonymous, so I left it as is.

agitter:
Thanks @SiminaB, it's merged.  The intro changes were successfully reverted and didn't show up in fc82312b516928948f54c666eed84c42e8c5b585.

Issue #680, opened 2017-11-03
Upgrading repo to track manubot-rootstock


dhimmel:
The Deep Review predates the Manubot (the CI-based system we use to process the manuscript). In fact, [`greenelab/manubot-rootstock`](https://github.com/greenelab/manubot-rootstock) (a template GitHub repo for Manubot instances) and [`greenelab/manubot`](https://github.com/greenelab/manubot) (a python package for processing the manuscript) were both derived from the code from Deep Review. In other words, Deep Review gave birth to these projects.

However, the Manubot has now diverged considerably from the codebase and architecture present in this repository. The manuscript processing code in this repository has not been receiving any rootstock updates. As time goes on, the reconciliation of the codebases becomes more difficult. However, once synchronized, incremental updates should generally be straightforward.

In https://github.com/greenelab/deep-review/issues/678#issuecomment-341539497, I proposed merging in the latest manubot-rootstock. @agitter [replied](https://github.com/greenelab/deep-review/issues/678#issuecomment-341612892):

> @dhimmel maybe we should open a new issue to preview and discuss how much would be involved in the merge. I'm hesitant to create a lot of new work when the current version of deep review is running smoothly even if it lacks some (important) features.

The merge certainly could be difficult. I'm happy to take on the task with a focus on minimizing disruption to open pull requests. The main benefits I see to the upgrade are:

+ getting new Manubot features including improved CI build feedback, formatting, and architecture
+ providing users with an up-to-date Manubot experience. The Deep Review is the Manubot instance with the greatest number of contributors. So we should strive to give users the cutting edge Manubot experience.
+ this repository has some customizations such as a non-standard (in terms of Manubot) authorship list. I'm interesting in seeing how the current Manubot architecture accommodates these needs. In other words, the Deep Review can provide important feedback for how to improve Manubot... but only if it's actually running Manubot.

Anyways, I'll attempt a merge and PR, so we can see exactly what the upgrade will entail. Most likely we will have to remove the current author scripts and re-engineer them, which I am willing to do.

agitter:
Good luck with the merge.  This will also add proper figure and table support, which we'll need for the revisions.

How much has changed from the author perspective?  The citation syntax is slightly different now.  Is that it?  We already asked all of the contributors learn one custom writing platform so I want to be conscious of how many new requirements we impose.

dhimmel:
> How much has changed from the author perspective?

[...]

PR #681, opened 2017-11-03
Update the Deep Review with latest Manubot

dhimmel:
Closes https://github.com/greenelab/deep-review/issues/680


dhimmel:
You can browse what the repo will look like if this PR is merged at [this tree](https://github.com/dhimmel/deep-review/tree/12ce17cdfa572cca61f16457fa99e705826719c9).

Here's the PDF from a local build: [`manuscript.pdf`](https://github.com/greenelab/deep-review/files/1442331/manuscript.pdf).

Before merging remember to create `output` branch to replace `references` branch.

@agitter ready for review

agitter:
@dhimmel I should have time to review this weekend

dhimmel:
> I should have time to review this weekend

@agitter if okay with you, I recommend we merge now, since I anticipate people may begin making revisions this weekend and I don't want them to start off of the old repository.

I could then address any issues you identify in subsequent commits.

agitter:
@dhimmel I can go through quickly now

agitter:
@dhimmel should I create a new issue to track the items above that need to be fixed or double-checked?

dhimmel:
> should I create a new issue to track the items above that need to be fixed or double-checked?

Yep, and you can assign me the ones you want me to do!

Some quick notes:

> Add back the example tag in USAGE.md

This seems like it should be an upstream manubot-rootstock change?

> Perhaps rename manuscript.pdf back to deep-review.pdf along with the other outputs

We could create a symlink pointing to `manuscript.pdf`. I'd only do this if we think there are public links (outside of this repo) to deep-review.pdf.

The arXiv build warnings are because those studies have now been published with DOIs. We can fix these at the end or sooner?

> Make a note to all authors about the changes such as the ; between multiple references and prohibited characters in citations

This should be partly done in the updated README.

agitter:
> Add back the example tag in USAGE.md

We had a Deep Review-specific example, `[@tag:Zhou2015_deep_sea]`, that I had in mind.  Any example from `citation-tags.tsv` would be fine, in which case it wouldn't be in the upstream manubot-rootstock.

> Perhaps rename manuscript.pdf back to deep-review.pdf along with the other outputs

We can leave this alone.  It was only my very minor personal preference.

> The arXiv build warnings are because those studies have now been published with DOIs.

We could fix them now or immediately before resubmitting. I expect other preprints could be published.  I just don't want to forget to do it.

> This should be partly done in the updated README.

Your README update took care of the semi-colon.  I'll use my #696 for the forbidden tags.

PR #707, opened 2017-11-16
suggest adding a word cloud figure in intro section

luzh2:


agitter:
Interesting idea.  What do we want the reader to learn from this figure?  It's pretty, but I don't see how it contributes to our message.

agitter:
@luzh2 I'm going to close this unless you have more thoughts on how this figure contributes to our message.

luzh2:
Dear Anthony,

Sorry that I missed your comment earlier. Please feel free to close this
specific one. I was hoping to generate some figure that could be seen as an
overview of the entire article; something that people could use in their
presentation. We created one for the bionlp section. Will have to think
more on this general one. Thanks!

Zhiyong

agitter:
Thanks @luzh2, and we do still plan to review your NLP section #567.

PR #717, opened 2017-12-13
LSTM now defined; remove extraneous definition

cgreene:
@SiminaB defined LSTM in the glossary table from #588 . I remembered that it was redefined here. This is a quick fix to get back to the intended "define in one place" mantra.


michaelmhoffman:
I think it would be best to define things on their initial occurrence in the main text *and* in the glossary table.

cgreene:
@michaelmhoffman works for me! :) Will close w/o merge 👍

agitter:
@dhimmel we are safe to delete the `lstm-now-defined` branch per this discussion.  I'm deleting now.

PR #721, opened 2017-12-16
Condense NLP section

agitter:
Following up on #567, I tried to condense some of the NLP section in categorize without losing any of the main points.  There were also several inconsistent acronyms.  @luzh2 can you please review to make sure the edited version is still correct?

After we merge this, we need to fix the clinical EHR sections.  We now have an EHR paragraph in the text mining sub-section and a text mining paragraph in the following EHR sub-section.  They mention some of the same papers.  This should be collapsed, and we need a better transition from the text mining to EHR sub-sections.

### DIFF
--- a/content/02.intro.md
+++ b/content/02.intro.md
@@ -27,13 +27,13 @@ Deep learning approaches grew from research in neural networks, which were first
 for how our brains process information. The history of neural networks is
 interesting in its own right [@doi:10.1103/RevModPhys.34.135]. In neural
 networks, inputs are fed into a hidden layer, which feeds into one or more
-hidden layers, which eventually link to an output layer. 
+hidden layers, which eventually link to an output layer.
 A layer consists of a set of nodes, usually called "features" or "units,"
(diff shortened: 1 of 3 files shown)


agitter:
A reference was dropped from the intro in #566.  I didn't want to make a separate pull request just for that so I'm adding it here.

luzh2:
i have reviewed the edits on the NLP subsection and they look good to me. thanks

agitter:
I'm going to merge this now that @luzh2 approved.

@luzh2 will you please be able to make another pull request to eliminate the redundant discussion between this section and the `Electronic health record` section?  There is too much overlap, but I don't have the right domain knowledge to merge these two sections.

yfpeng:
Sure. I will work with @luzh2 to address this issue.

agitter:
Thanks @luzh2 and @yfpeng.  I don't think it will be too hard.  Mostly a matter of deciding which section to delete the redundant text and references from.

Issue #561, opened 2017-07-05
Authors since initial submission


cgreene:
I wanted to track this so that we don't lose contributors.

Tagging @avctrh 

New authors:
- [x] @austinvhuang (formerly @avctrh)
- [x] @SiminaB 
- [x] @michaelmhoffman 
- [x] @luzh2 
- [x] @lavenderca
- [x] @zietzm
  
Acknowledgement -> Author due to approval after v1 went out:
- [x] @swamidass

agitter:
Perfect.  I'll add new authors in one batch once we are nearing a new version.

agitter:
I added @SiminaB to the list above after merging #588.  That pull request alone may not meet ICMJE authorship criteria, but the discussion in #565 and incoming #566 do.

agitter:
I added @michaelmhoffman after merging #620, noting that several other pull requests are still incoming.

michaelmhoffman:
Thanks! Please don't forget to remove me from acknowledgments.

evancofer:
This may be a silly question to ask on this pull request, but [how] do we plan on updating the author ordering? If we are still using the randomized hierarchies, are any transitions between them expected?

agitter:
@evancofer It's a good idea to discuss this.  The algorithm I proposed in #369, [which is demoed here](https://github.com/agitter/deep-review/blob/author-order/authors/author_order_demo.ipynb), would:
- keep the random order for existing authors within each category
- randomly order the new authors within each category
- append the new authors to the end of the existing author list within each category
- concatenate the category-specific author lists

The main objectives were to continue to emphasize the categories instead of the specific order and also minimize changes.  I thought it would be confusing to have major permutations to the author list every time we add someone new.

michaelmhoffman:
As a new author I think it is only fair that we get put at the end of whatever category we end up in.

agitter:
Added @luzh2 above after merging #567

agitter:
Added @lavenderca after merging #629

agitter:
Added @zietzm after merging #638

cgreene:
Talked with @swamidass at PSB. He contributed to the discussion of issues + writing on the first round (see Acknowledgements). He is now able to approve. I'm not sure what our procedure is this time around, but we should include him when we have things ready.

swamidass:
Yes, I do approve.

agitter:
@swamidass great, we'll add you in the next version.  I'll post instructions for adding yourself here.

[...]

PR #749, opened 2018-01-06
New author template

agitter:
I'm adding a temporary authors template.  The new authors in #561 can add their information here. 
 Then I'll randomly insert them into the order and `metadata.yaml` using the previous [algorithm](https://github.com/greenelab/deep-review/blob/master/authors/author-order.ipynb).  I dropped the `twitter` key-value pair because we aren't using that in this manuscript and left two example authors from manubot-rootstock.

Also fixing a funding error from our previous author table.


cgreene:
This looks good to me 👍

PR #765, opened 2018-01-11
Small fix

SiminaB:
Inputs are fed into input layer, not into hidden layer.


SiminaB:
I think it may have been my own typo before actually. :-)

Issue #700, opened 2017-11-10
Post-Manubot update cleanup


agitter:
We wanted to merge #681 quickly, so here are some lingering questions and confirmations from that merge:
- [ ] Inspect pdf formatting and references in detail
- [ ] Ensure that all author info transferred correctly
- [ ] Customize the front matter to reduce the length of the author list
- [ ] Add back footnotes for corresponding authors and the randomized author order
- [ ] Replace contribution, funding, and other templates

@dhimmel if you are still willing to re-engineer the author template scripts, that will address most of these.  I'll review the formatting and author info that I didn't scrutinize during the #681 review.

agitter:
While reviewing the author information I noticed the unnecessary `OTHER` in Yifan Peng's funding information.  I'll leave it until we decide what we're going with the templates.

https://doi.org/10.1371/journal.pbio.1001661 isn't working with the new CSL.  The last author is not show, perhaps because it is a consortium.  The rest of the pdf formatting and reference list looked good.

agitter:
@dhimmel do you still plan to work on the new templating that will replace the author table processing functions?  We can at least decide which templates that were lost in #681 should be re-implemented.

dhimmel:
> do you still plan to work on the new templating that will replace the author table processing functions?

Yeah. It seems like that the author section is the only part that still needs reworking? I could make it like the format from [preprint v1](https://www.biorxiv.org/content/early/2017/05/28/142760.full.pdf), if that's what you prefer. Let me know if there are any improvements on that format you'd like before I get started.

agitter:
I believe daef821 is the last time I added templates.  There I see templates in the [front matter](https://github.com/greenelab/deep-review/blob/daef821cc87f264063cb0f32e79d284a756d56bc/sections/00_title_authors.md) for listing authors, affiliations, and corresponding authors and the [methods](https://github.com/greenelab/deep-review/blob/daef821cc87f264063cb0f32e79d284a756d56bc/sections/08_methods.md) for funding.  I thought I also had them for the contributions statement, but apparently I didn't implement that.

With so many authors, I do prefer having the front matter author list be more compact.  It wouldn't need to be identical to `preprint v1`, but something that puts all authors in a single wrapped line followed by a list of affiliations will fit better on a single PDF page.  We could make the affiliations smaller font to further save space.  The notes about author order and corresponding authors could be hard coded because those don't change.  We do have some repeated affiliations.

[...]

PR #769, opened 2018-01-11
Frontmatter: use numbered affiliations

dhimmel:
Switch frontmatter to condensed format which uses numbered affiliations.

Refs https://github.com/greenelab/deep-review/issues/700

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -18,6 +18,9 @@ on {{date}}.
 ## Authors
 
 {% for author in authors %}
+{% if author.orcid is defined -%}
+[![ORCID icon](images/orcid.svg){height="11px" width="11px"}](https://orcid.org/{{author.orcid}})
+{%- endif %}
 {{author.name}}<sup>{{author.affiliation_numbers | join(',')}}</sup>
 {%- if not loop.last -%}, {%- endif -%}
 {% endfor %}
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -21,9 +21,12 @@ on {{date}}.
 {% if author.orcid is defined -%}
 [![ORCID icon](images/orcid.svg){height="11px" width="11px"}](https://orcid.org/{{author.orcid}})
 {%- endif %}
-{{author.name}}<sup>{{author.affiliation_numbers | join(',')}}</sup>
+{{author.name}}<sup>{{author.affiliation_numbers | join(',')}}{%- if author.symbol_str is defined -%},{{author.symbols | join(',')}}{%- endif -%}</sup>
 {%- if not loop.last -%}, {%- endif -%}
 {% endfor %}
+
+<sup>☯</sup> --- Author order was determined with a randomized algorithm<br>
+<sup>†</sup> --- To whom correspondence should be addressed: `gitter` at `biostat.wisc.edu` (AG) and `csgreene` at `upenn.edu` (CSG)
 <small>
 
 {% for affiliation in affiliations %}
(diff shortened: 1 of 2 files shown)


dhimmel:
I tried a version that hyperlinked the ORCID and GitHub icons for each author using the following jinja2/md:

```
{%- if author.orcid is defined -%}
[![ORCID icon](images/orcid.svg){height="10px" width="10px"}](https://orcid.org/{{author.orcid}})
{%- endif -%}
{%- if author.github is defined -%}
[![GitHub icon](images/github.svg){height="10px" width="10px"}](https://github.com/{{author.github}})
{%- endif -%}
{%- if author.twitter is defined -%}
[![Twitter icon](images/twitter.svg){height="10px" width="10px"}](https://twitter.com/{{author.twitter}})
{%- endif -%}
```

I didn't love the look:

screenshot from 2018-01-11 15-11-21

What do others think.

I also considered hyperlinking authors to ORCIDs, but then the names are blue.

dhimmel:
@agitter I was thinking we'd add additional symbols as superscripts following the affiliation numbers. Who should have what superscripts?

agitter:
Maybe the icons would look less crowded if you include ORCID but not GitHub?  That's what [bioRxiv does](https://doi.org/10.1101/142760) and it looks nice there.  They also have the icon precede the author name, which I'm indifferent about.

The uploaded manuscript was a manubot-rootstock build.

I think we can use the same superscripts as last time unless anyone has other suggestions:
- Travers Ching <sup>*</sup>
- Anthony Gitter <sup>†</sup>
-  Casey S. Greene <sup>†</sup>

<sup>*</sup> Author order was determined with a randomized algorithm

<sup>†</sup> To whom correspondence should be addressed: gitter@biostat.wisc.edu (AG) and csgreene@upenn.edu (CSG)

dhimmel:
I added ORCID icons before author names in https://github.com/greenelab/deep-review/pull/769/commits/fa6432e9d77e1f16b1a6f3298e30532d721388bf. Unfortunately, I couldn't get nonbreaking spaces to work so some author names are on different lines than their ORCID icon.

I added the symbols in https://github.com/greenelab/deep-review/pull/769/commits/8484b3472a8b5ec04aee659ace3e9ac430611aaf. Note that I had to remove the `@` symbols due to https://github.com/greenelab/manubot/issues/25. I also switched the authorship symbol to something slightly more traditional: ☯ as opposed to *. Can change back if others do not like.

Here's the latest look:

authors

content/00.front-matter.md
```diff
@@ -21,9 +21,12 @@ on {{date}}.
 {% if author.orcid is defined -%}
 [![ORCID icon](images/orcid.svg){height="11px" width="11px"}](https://orcid.org/{{author.orcid}})
 {%- endif %}
-{{author.name}}<sup>{{author.affiliation_numbers | join(',')}}</sup>
+{{author.name}}<sup>{{author.affiliation_numbers | join(',')}}{%- if author.symbol_str is defined -%},{{author.symbols | join(',')}}{%- endif -%}</sup>
 {%- if not loop.last -%}, {%- endif -%}
 {% endfor %}
+
+<sup>☯</sup> --- Author order was determined with a randomized algorithm<br>
```
agitter: We can change ☯ back to * later if someone objects.  The way you implemented the superscripts is easy to change.

content/00.front-matter.md
```diff
@@ -21,9 +21,12 @@ on {{date}}.
...
+<sup>†</sup> --- To whom correspondence should be addressed: `gitter` at `biostat.wisc.edu` (AG) and `csgreene` at `upenn.edu` (CSG)
```
agitter: This workaround looks good to me.  We could use `gitter at biostat.wisc.edu` instead of `gitter` at `biostat.wisc.edu`, but I don't have a strong preference.

PR #772, opened 2018-01-12
Split affiliations and correct typo

agitter:
Now that we have better support for multiple affiliations, I'm splitting @akundaje's.

This will avoid having `Department of Computer Science, Stanford University, Stanford, CA` and `Department of Genetics and Department of Computer Science, Stanford University, Stanford, CA` as separate affiliations.


agitter:
Noticed an undefined abbrevaition and piggybacking on this pull request to fix it.

Issue #689, opened 2017-11-03
Referee 2.2


cgreene:
> There is a slight imbalance in the presentation of various application areas. The section on drug development (p40ff), for example, is quite extensive and provides a lot of technical details which might be less relevant for a reader who tries to get a general overview of deep learning in biomedical research. An area which is little mentioned on the other hand are deep learning approaches to brain data, eg connectivity maps, and the area of learning from structured data, such as graphs.

agitter:
Should I cut some of the technical details in the drug discovery section?

Structured data, specifically graphs, seems in-scope to me.  We could add a short section about graphs, PPI networks, etc.  @zietzm does that still interested you?  A graph analysis section could add to #638.  We have #241 #543 and perhaps other issues I'm forgetting.  This section could be connected to the graph-based methods we discuss in the drug discovery section.

I'm less sure about brain data.  There have been many applications of deep learning in neuroscience.  A short paragraph may not due them justice, and it is debatable whether it is in the scope of this review.

zietzm:
Absolutely, I will get on that right away.

agitter:
@srinituraga agreed to help us with a couple paragraphs on deep learning for neuroscience and connectivity maps!

cgreene:
Wonderful! :)

agitter:
> The section on drug development (p40ff), for example, is quite extensive and provides a lot of technical details which might be less relevant for a reader who tries to get a general overview of deep learning in biomedical research.

@cgreene I can make revisions to address this comment.  Do you think I should trim this section?  Remove detail but leave the same references?

agitter:
#638 added methods for learning on graphs.

cgreene:
From a read through, the section on ligand-based prediction is one of the longest. However, the section itself on treatments is not that long. What do you think about splitting the representation learning portion at `Much of the recent excitement in this domain has come from what could be considered a creative experimentation phase, in which deep learning has offered novel possibilities for feature representation and modeling of chemical compounds` into its own subsection?

agitter:
@cgreene I could split the content, but I'm also wondering whether I would also need to make cuts to shorten the overall length or make it less technical.

cgreene:
I didn't see anything that stuck out as glaringly obviously overly technical. I found that on a read through it's longer than other areas because it deals with more topics. However, if one considers the part about learning a representation of chemicals as a separate topic, then things are a bit more in balance.

[...]

Issue #654, opened 2017-09-08
Did Kaggle Predict Drug Candidate Activities? Or Not?


agitter:
[A comment on bioRxiv](http://www.biorxiv.org/content/early/2017/05/28/142760#comment-3444535593) pointed out criticism of the Merck Kaggle competition, referencing Anthony Nicholls's comments specifically.  Those may be the comments in [this Derek Lowe article](http://blogs.sciencemag.org/pipeline/archives/2012/12/11/did_kaggle_predict_drug_candidate_activities_or_not).

This issue is a reminder to look into those comments and assess whether we are presenting the Merck competition fairly in the `Ligand-based prediction of bioactivity` section.

agitter:
Closed by #774

PR #638, opened 2017-08-10
Added PPI section with MHC subsection

zietzm:
References #575 and includes [MHC-peptide papers](https://github.com/greenelab/deep-review/issues?utf8=%E2%9C%93&q=MHC)

Added a section on Protein-Protein Interactions (PPI) with a subsection on MHC-peptide binding prediction.

@agitter mentioned PPI networks as a possible area of interest in #575, but this has been neglected here for the sake of not adding too much to an already long section. If a PPI network subsection is still desired, I would be more than happy to add one, but I understand the necessity to minimize additional length being added to this paper.

Issue #773, opened 2018-01-13
Covariant Compositional Networks For Learning Graphs


agitter:
https://arxiv.org/abs/1801.02144

> Most existing neural networks for learning graphs address permutation invariance by conceiving of the network as a message passing scheme, where each node sums the feature vectors coming from its neighbors. We argue that this imposes a limitation on their representation power, and instead propose a new general architecture for representing objects consisting of a hierarchy of parts, which we call Covariant Compositional Networks (CCNs). Here, covariance means that the activation of each neuron must transform in a specific way under permutations, similarly to steerability in CNNs. We achieve covariance by making each activation transform according to a tensor representation of the permutation group, and derive the corresponding tensor aggregation rules that each neuron must implement. Experiments show that CCNs can outperform competing methods on standard graph learning benchmarks.

See the [ICLR 2018 discussion](https://openreview.net/forum?id=S1TgE7WR-) as well.

PR #774, opened 2018-01-13
Drug discovery and chemical representation

agitter:
Work in progress.  Ref #689.

I plan to add a few new references on graph convolutions (without much discussion) to complement our new section on PPI graphs in #638.  This also adds a reference to partially address the [bioRxiv comment](http://www.biorxiv.org/content/early/2017/05/28/142760#comment-3444535593) per #654.


agitter:
@dhimmel ef2b76c6612aa68c741783c5388e420bef21acad passed the build yesterday but failed today when I reran it to try to diagnose my more recent build failures.

Have you seen this error before:
```
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/manubot/bin/pandoc-fignos", line 11, in <module>
    sys.exit(main())
  File "/home/travis/miniconda/envs/manubot/lib/python3.6/site-packages/pandoc_fignos.py", line 415, in main
    plusname, starname, 'figure')
TypeError: replace_refs_factory() missing 1 required positional argument: 'target'
Error running filter pandoc-fignos:
Filter returned error status 1
```

agitter:
This may need some additional work but is ready for feedback and review.  I didn't go into detail about the new graph methods because the reviewer already thinks this section is too technical.  I primarily sought to separate the representation learning section to break up the long ligand-based subsection and emphasize the graph-based methods.  These were requested by the same reviewer and are interesting algorithms beyond the virtual screening domain.

I could cut some text if we'd like to shorten this.

This PR adds #555, #641, #741, #762, #773, and #776.

dhimmel:
Nothing stands out to me as problematic in this PR's diff.

@tomduck (creator of `pandoc-fignos`), any idea what [this error](https://travis-ci.org/greenelab/deep-review/builds/328724475#L775) could be caused by:

```
TypeError: replace_refs_factory() missing 1 required positional argument: 'target'
Error running filter pandoc-fignos:
```

We recently upgraded pandoc-fignos, pandoc-tablenos, and pandoc-egnos to v1.0 in https://github.com/greenelab/deep-review/commit/6c2020f8b8fef9c72dba122ecf378499e3db3c53. @agitter do you know whether the successful builds used the older version of `pandoc-fignos`?

tomduck:
The API for pandoc-xnos was slightly changed.  You need to update it as well.

tomduck:
Actually, it looks like you may be using the new pandoc-xnos but an old pandoc-fignos.  Can you upgrade both to the most recent version?

Note that I am in the process of preparing another release, which will go out within the next hour.

dhimmel:
Looks like our [build](https://travis-ci.org/greenelab/deep-review/builds/328724475#L615) used `pandoc-eqnos==1.0.0`, `pandoc-fignos==1.0.0`, `pandoc-tablenos==1.0.0`, and `pandoc-xnos==0.15`.

> Actually, it looks like you may be using the new pandoc-xnos but an old pandoc-fignos. Can you upgrade both to the most recent version?

Will do soon, after the next release is out. I think the issue may be that we pinned fignos, egnos, and tablenos, but not xnos. Therefore, the error would have first occurred when a higher version pandoc-xnos started getting installed.

tomduck:
Yes, that makes sense @dhimmel.  Is there something I should be doing on my end to prevent such a problem in the future?

dhimmel:
As per [here](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies), I think you could specify something like `pandoc-xnos>=0.13,<0.15`, in [`setup.py`](https://github.com/tomduck/pandoc-fignos/blob/847125228bd2ec8dc0c2ad08804415da81a64b1f/setup.py#L111) of pandoc-fignos. However, I'm not sure how you would know, when making pandoc-fignos==1.0 that it would stop working at pandoc-xnos==0.15, which presumably hadn't been released yet.

tomduck:
Yes, that is exactly the problem.  I will think on it and come up with a solution.

content/05.treat.md
```diff
@@ -165,7 +168,7 @@ When protein structure is available, virtual screening has traditionally relied
 Recently, deep learning approaches have been developed to model protein structure, which is expected to improve upon the simpler drug-target interaction algorithms described above that represent proteins with feature vectors derived from amino acid sequences [@doi:10.1109/BIBM.2014.6999129; @doi:10.1021/acs.jproteome.6b00618].
 
 Structure-based deep learning methods differ in whether they use experimentally-derived or predicted ligand-target complexes and how they represent the 3D structure.
-The Atomic CNN [@arxiv:1703.10603] takes 3D crystal structures from PDBBind [@doi:10.1021/jm048957q] as input, ensuring it uses a reliable ligand-target complex.
+The Atomic CNN [@arxiv:1703.10603] and TopologyNet [@doi:10.1371/journal.pcbi.1005690] take 3D structures from PDBBind [@doi:10.1021/jm048957q] as input, ensuring the ligand-target complexes are reliable.
```
cgreene: I think you need to either remove `The` or add `networks` or `models`.

agitter:
> do you know whether the successful builds used the older version of pandoc-fignos

@dhimmel I didn't explicitly change anything in our environment between when the build passed and failed.  However, our [last successful build](https://travis-ci.org/greenelab/deep-review/builds/328442851#L1725) on master did use `pandoc-xnos` version 0.14.

It looks like we have a root cause now and can merge this once we pull in greenelab/manubot-rootstock#99  Thanks for the help @tomduck.

dhimmel:
@agitter, I just merged https://github.com/greenelab/deep-review/pull/778 so you should be good to go.

PR #777, opened 2018-01-14
Added author information for AH

austinvhuang:


austinvhuang:
@agitter  - added now, thanks for the note.

PR #779, opened 2018-01-14
Update email addresses

dhimmel:
@cgreene is this still the email address you'd like to use?


cgreene:
@dhimmel : thanks! good point :)

dhimmel:
@cgreene might as well also update the following in this PR:

https://github.com/greenelab/deep-review/blob/d22a3f8aa4767a36144a18e27e9cf1cd48443baa/content/metadata.yaml#L188

Issue #602, opened 2017-07-27
`F-measure` vs. `F1` vs. F_1


michaelmhoffman:
The paper uses both `F1` and `F-measure` (which I assume specifically means F1) in different places. I would suggest changing all to F_1 with a subscript 1.

michaelmhoffman:
I checked that the paper cited for F-measure is specifically using F1.

PR #789, opened 2018-01-15
`F1`/`F-measure` -> `F~1~ score`

michaelmhoffman:
Addresses greenelab/deep-review#602.


agitter:
@dhimmel do we know how pandoc will render this?  Github flavored Markdown coverts it to strikethrough: F~1~

dhimmel:
Probably not as intended. I'd suggest using the unicode subscript 1, like F₁ (copy and paste it). You could also do `<sub>1</sub>`.

michaelmhoffman:
I didn't read this section of the Pandoc manual carefully enough. Looks like you may need the `subscript` extension enabled for this to work. That it breaks the Github-flavored Markdown is bad. Will try Unicode subscript 1.

https://pandoc.org/MANUAL.html#superscripts-and-subscripts

agitter:
I didn't build the manuscript locally to test it, but I'll fix this if the Travis build looks bad after merging.

agitter:
The build finally finished, and the Unicode subscript 1 works.

PR #804, opened 2018-01-18
Fix broken Gal reference

agitter:
Showed up as
`432. Yarin Gal - Publications | Oxford Machine Learning http://www.cs.ox.ac.uk/people/yarin.gal/website/publications.html`

This may require a manual override, but I'm trying this first.


agitter:
Greycite couldn't handle the PDF URL.  I'll update this with a manual reference.

agitter:
Much better with the manual reference.  Here is my local build:
image

@cgreene ready for review.

PR #799, opened 2018-01-17
Draft Response to Reviewers

cgreene:
Initial steps in this area. Currently WIP. @agitter feel free to pick up if you get time.

PR #806, opened 2018-01-19
Adjust relative column width in Table 1

agitter:
I need to revert some of the table column width changes from #799.  The table doesn't look very good.
image


agitter:
It's much better now:
image

PR #817, opened 2018-02-16
Update Manubot on 2018-02-16

dhimmel:


agitter:
@dhimmel what do you think about creating the versioned webpage directories retroactively for deep review?

dhimmel:
> what do you think about creating the versioned webpage directories retroactively for deep review?

That's the plan

agitter:
@dhimmel is this [Travis error](https://travis-ci.org/greenelab/deep-review/builds/342500448#L1928) something to be concerned about?
```
Checkout returned a nonzero exit status. See stderr:
error: pathspec 'v' did not match any file(s) known to git.
```

It looks like we did get the [versioned webpage directories](https://github.com/greenelab/deep-review/tree/gh-pages/v) and the [freeze](https://greenelab.github.io/deep-review/v/freeze/), and [latest](https://greenelab.github.io/deep-review/v/latest/) URLs work

dhimmel:
That error will occur only for the first time continuous deployment is used with the new setup. That's because the previous gh-pages commit does not contain a `v` directory.

PR #830, opened 2018-03-05
Remove newline

agitter:
The two trailing spaces are treated as a line break
image

I haven't systematically scanned the markdown files to see if we have this elsewhere.


agitter:
I found other pairs of spaces, but not at the ends of lines.

PR #951, opened 2019-03-22
PR to update Manubot to rootstock from 2019-03-21

dhimmel:
Includes new interactive frontend.


build/plugins/analytics.html
```diff
@@ -9,3 +10,5 @@
   ga('send', 'pageview');
 </script>
 <!-- End Google Analytics -->
+
+<!-- copy and paste code from Google Analytics or similar service here -->
```
agitter: Is this the intended code placement?  The Google Analytics code that was here is above the comment.  It probably doesn't matter enough to change.

ci/deploy.sh
```diff
@@ -48,8 +49,8 @@ This build is based on
 https://github.com/$TRAVIS_REPO_SLUG/commit/$TRAVIS_COMMIT.
 
 This commit was created by the following Travis CI build and job:
-https://travis-ci.org/$TRAVIS_REPO_SLUG/builds/$TRAVIS_BUILD_ID
-https://travis-ci.org/$TRAVIS_REPO_SLUG/jobs/$TRAVIS_JOB_ID
+https://travis-ci.com/$TRAVIS_REPO_SLUG/builds/$TRAVIS_BUILD_ID
```
agitter: What do we want to do with manuscripts still on travis-ci.org?  Should we leave this?

dhimmel: I will revert this change. Manuscripts still on travis-ci.org should still use the .org link. However, since it hasn't been a conflict, I've missed it in my previous upgrades.

build/plugins/analytics.html
dhimmel: Yeah I saw that. I think `analytics.html` maybe should be updated so there is no blank line between the two comments and there is a final comment like `<!-- palce analytics above`

agitter: Okay.  So it's an upstream change and nothing we need to do here.

agitter:
@dhimmel the updated deep review looks great.  The improved citation features are a huge help here.

[Figure 1](https://greenelab.github.io/deep-review/#fig:nn-petting-zoo) and [Figure 2](https://greenelab.github.io/deep-review/#fig:biotm) do not display well when expanded because they do not have a background color. Do we need to change the lightbox usage?

A few paragraphs above https://greenelab.github.io/deep-review/#electronic-health-records I was surprised that `[96]` is split across two lines.  It happens again in the paragraph above https://greenelab.github.io/deep-review/#promoters-and-enhancers

https://greenelab.github.io/deep-review/#continuous-collaborative-manuscript-drafting suggests it may be a problem with links in general not wrapping as expected, not just citation links.

agitter:
Regarding line wrapping, see also `pandoc-citeproc` here
image

dhimmel:
A quick scan and I don't see any words split across lines that are not in links. Is that so?

vincerubinetti:
I thought you were talking about breaking in between brackets with multiple citations, ie 
```
[1, 2,    
3]
```

Yes, I made the links break more easily. I can't remember exactly why I did this, but it was probably due to problems with long links in tables, or just making it look better when really long links took up the whole page width. You can see what it's like with that behavior disabled by doing this:

image

dhimmel:
> probably due to problems with long links in tables, or just making it look better when really long links took up the whole page width.

Yeah I can see why for long links this should be the behavior. But it's definitely odd for hyperlinks on words or phrases. I wonder if there is any way to apply this behavior when the display text is a URL but not when the display text is not a URL.

vincerubinetti:
To clarify on this, there is no CSS way to apply a certain style to an element based on the text it contains. In other words, if you want to wrap based on the length of the link text, or try to wrap if the link text appears to be a URL and not just a sentence, that will require javascript.

Issue #957, opened 2019-06-21
Table of contents for paper


AnneCarpenter:
It'd be lovely to generate a table of contents for the full paper, so people can see an overview of the topic!

evancofer:
@AnneCarpenter I agree! @dhimmel @cgreene is this something that can be easily added with Manubot?

dhimmel:
The [HTML view](https://greenelab.github.io/deep-review/v/ec05a06890580987fd84eeadaaa8422bda2f918f/) has a slide-out table of contents:

> ![image](https://user-images.githubusercontent.com/1117703/59964655-5c5a2c80-94d1-11e9-974f-352194cfc2a4.png)

If you update to the latest Manubot, the default behavior of the interactive HTML TOC will be a bit more intuitive. @vincerubinetti the deep review seems to be an example where the TOC headings are a bit large in the side bar?

@AnneCarpenter and @evancofer is this what you're looking for? If so, there is an option to make the side bar TOC always open, if that would help make users aware of it's existence.

If you would like a TOC to be included in the actual document as a section, including in the PDF, then you could try adding Pandoc's `--table-of-contents` argument in `build.sh`. Haven't ever used this, but would be interested in seeing it.

AnneCarpenter:
That solves my problem!! Indeed, setting the sidebar as default would avoid people like me missing it (the downside would be that sometimes folks are annoyed when there is a lot of clutter on the page and they just want to see the content!)

agitter:
Thanks for the suggestion @AnneCarpenter !

I haven't tried adding `--table-of-contents` to the PDF build like @dhimmel suggested, but the current PDF does show bookmarks correctly as an alternative overview:
![image](https://user-images.githubusercontent.com/3532898/59968248-604f7400-94fc-11e9-9d21-b17cf768fbc9.png)

evancofer:
@dhimmel Yes, I think it looks good. I agree with @AnneCarpenter that the expanded-by-default behavior could be nice, but I do think it could get in the way on smaller mobile screens. That being said, I think a paper of this size may be destined for reading difficulties on mobile viewers.

dhimmel:
> the current PDF does show bookmarks correctly as an alternative overview

[...]

PR #958, opened 2019-06-24
Upgrade rootstock on 2019-06-24

dhimmel:
Includes improved default behavior for the TOC plugin related to https://github.com/greenelab/deep-review/issues/957.


dhimmel:
Build failed due to some intermittent error with Crossref. Reminder to rerun it https://travis-ci.org/greenelab/deep-review/builds/550336352

Do we want to freeze the reference metadata for Deep Review?

agitter:
> Do we want to freeze the reference metadata for Deep Review?

If we do that, we may need to review all the references to make sure the current metadata we are pulling is accurate.  I did that around the time of submission but it may have diverged since then.

Issue #684, opened 2017-11-03
Referee 1.2


cgreene:
> The authors discussed deep learning models such as sDA, CNN, RNN etc. It would be better to have a figure illustrating their architectures. This way, the reader will have a concrete visualization that will aid the understanding of the discussion points in the manuscript.

agitter:
Related to #354 #566 #630

cgreene:
I am starting to work on this. I've tried to get a relatively modestly sized figure that covers the main architectures that we discuss in the text.

Here's a first pass at something that covers a few of the architectures from @SiminaB's glossary.
![nn-petting-zoo](https://user-images.githubusercontent.com/542643/34794833-eef28974-f61d-11e7-9df6-78c54b97ba90.png)

This is in a google draw page, as @evancofer suggested:
https://docs.google.com/drawings/d/1TEzlqiCewmJ-EITvogMqDz7N6CizMtBvWE4H_BXqnuQ/edit

Currently it is marked world editable.

cgreene:
BTW these schematics are heavily derived from http://www.asimovinstitute.org/neural-network-zoo/ so we definitely should discuss Fjodor in the legend. That's why I called it the "petting zoo" in the google draw title.

cgreene:
Oh, wait, my deep RNN is broken 😜 . Too focused on making the recurrent bits and i missed the output layer!

SiminaB:
I added edges to the legend as well - can remove or edit as desired!

SiminaB:
I'm wondering if we can add something to the figure about which types of networks are more used for supervised learning vs unsupervised learning. For example, if the outer layer matches the input layer, I guess that's always used for dimension reduction, so it would be unsupervised learning (although the reduced dimensions could then be used for supervised learning if there is an outcome). I don't have enough expertise to add this, but I think it would really help to clarify things for folks coming from stats (like myself).

cgreene:
Here's an updated version of the figure.
![nn-petting-zoo 1](https://user-images.githubusercontent.com/542643/34907749-3115860a-f851-11e7-99a7-03d4d8a27802.png)

cgreene:
Updating the figure so that I can link it in the new version of PR #775. 
![nn-petting-zoo 2](https://user-images.githubusercontent.com/542643/34995383-52363dc0-faa4-11e7-8cdd-47966c1a2a0f.png)

[...]

PR #978, opened 2019-09-05
petting-zoo figure: link to source

dhimmel:
link to source google drawing, which was previously difficult to track down.

CC @vincerubinetti 

Refs https://github.com/greenelab/deep-review/issues/684


dhimmel:
> did you test that the svg figure works as expected?

no. let me enable appveyor, since https://github.com/greenelab/deep-review/pull/977 updated rootstock with appveyor support.

agitter:
@dhimmel can you please restart the AppVeyor build so we can preview it?  I don't have permission.

PR #983, opened 2019-12-03
Make clear that update is in progress

cgreene:
There has been a fair bit of twitter traffic about the paper. This updates the title to try to make clear that the update is in progress and not yet finalized. Hopefully this will help readers understand this.

### DIFF
--- a/content/00.front-matter.md
+++ b/content/00.front-matter.md
@@ -5,7 +5,7 @@
 ##}
 
 _A published version of this manuscript from 04 April 2018 is available at <https://doi.org/10.1098/rsif.2017.0387>.
-An effort is underway to update the manuscript to capture the state of the field in 2019.
+An effort is underway to update the manuscript.
 New authors and links to new sections are available in [GitHub Issue #959](https://github.com/greenelab/deep-review/issues/959)._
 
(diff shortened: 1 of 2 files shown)


agitter:
Should we drop the "2019" from the title?  If we do release a new version, it definitely will not be this month!

stephenra:
@agitter Agree that "2019" ought to be dropped.

agitter:
We might as well drop 2019 from the header as well:
> An effort is underway to update the manuscript ~to capture the state of the field in 2019~.

evancofer:
Since there are only a few more weeks in the year, I agree that 2019 should probably be dropped.

content/00.front-matter.md
```diff
@@ -5,7 +5,7 @@
 ##}
 
 _A published version of this manuscript from 04 April 2018 is available at <https://doi.org/10.1098/rsif.2017.0387>.
-An effort is underway to update the manuscript to capture the state of the field in 2019.
```
agitter: @cgreene is it okay with you if I remove this part as well?

cgreene: Yes - thanks!

PR #997, opened 2020-03-06
Randomize author order on CI builds

dhimmel:
See discussion in https://github.com/manubot/manubot/pull/214

uses commit hash as random seed

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -11,6 +11,9 @@ export TZ=Etc/UTC
 # Default Python to read/write text files using UTF-8 encoding
 export LC_ALL=en_US.UTF-8
 
+# Randomize authors in metadata.yaml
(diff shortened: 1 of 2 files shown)


dhimmel:
--only-on-ci didn't work on GitHub. Will look into.

dhimmel:
talked with @cgreene and the current implementation is not exactly what is desired. The goal is to fix the order of the v1 authors and only shuffle the order of v2 authors.

Do we want v1 authors to appear individually in the metadata and author list? Or instead do we want to condense them all into an entity like "Deep Review v1 Authors"?

I think probably we should just hardcode / copy the rendered output as the v1 authors, since this is no longer something static.

cgreene:
I have hardcoded the v1 authors in #998. I think I need to merge this and then test that one. Will do that now.

PR #998, opened 2020-03-10
Hardcode v1 authors and remove them from metadata

cgreene:
I tried to pull from @dhimmel's branch associated with #997, make some more updates to better address the complexities, and then push back but I think I must have pushed to the wrong place.


cgreene:
Ok - I'm pretty sure this will work now.

dhimmel:
This doesn't contain the work from https://github.com/greenelab/deep-review/pull/997 yet, but that is OKAY. these two PRs can be merged independently.

I wonder if we should delete the v1 authors in metadata or move them to a field like `v1_authors`. Deletion is probably okay for now, we can always add them back in the future should we need them.

Looks like the newlines between affiliations have been removed. maybe because no longer a numbered list?

cgreene:
Ahh! I was wondering what happened with those newlines. You're right - making the affiliations have a version prefix broke the numbered list. I could make a bulleted list with the numbers after them, add an additional line (though that'd make a paragraph space), or something else. What do you think?

dhimmel:
Just read about pandoc's [`example_list`](https://pandoc.org/MANUAL.html#extension-example_lists) extension ([example](https://pandoc.org/try/?text=%0A%40key-1.+My+first+example+will+be+numbered+1.%0A%40key-2.+My+second+example...%0A%0ATake+a+look+at+%40key-2&from=markdown&to=plain)). Don't think we need it.

> I could make a bulleted list with the numbers after them, add an additional line (though that'd make a paragraph space), or something else. What do you think?

The easiest solution at the moment would be to add `  ` (two spaces) at the end of affiliation lines (same effect as adding `<br>`).

cgreene:
I had no idea that was possible!

content/00.front-matter.md
```diff
@@ -37,33 +37,136 @@ on {{manubot.date}}.
...
 
-<sup>♠</sup> --- Author order for version 2.0 is currently arbitrary.<br>
+<sup>♠</sup> --- Author order for version 2.0 is currently randomized [with each new build](https://github.com/greenelab/deep-review/pull/997).<br>
 <sup>†</sup> --- To whom correspondence should be addressed: greenescientist@gmail.com (C.S.G.)
 
+<small>
+
+{% for affiliation in manubot.affiliations %}
+2.{{affiliation.affiliation_number}}. {{affiliation.affiliation}}  
+{%- endfor %}
```
dhimmel: I think this will fix the lack of line breaks

```suggestion
{% endfor %}
```

content/00.front-matter.md
```diff
@@ -37,33 +37,136 @@ on {{manubot.date}}.
...
-{% for author in manubot.authors %}
-  {% if author.v1 -%}
-    {% if author.orcid is defined -%}
-    [![ORCID icon](images/orcid.svg){height="11px" width="11px"}](https://orcid.org/{{author.orcid}})
-    {%- endif %}
-    {{author.name}}<sup>{{author.affiliation_numbers | join(',')}}{%- if author.symbol_str is defined -%},{{author.symbol_str}}{%- endif -%}</sup>
-    {%- if not loop.last -%}, {%- endif -%}
-  {%- endif -%}
+{% for affiliation in manubot.affiliations %}
+2.{{affiliation.affiliation_number}}. {{affiliation.affiliation}}
```
dhimmel: Now that we have the newlines we need to remove these two spaces

```suggestion
2.{{affiliation.affiliation_number}}. {{affiliation.affiliation}}
```

content/00.front-matter.md
```diff
@@ -37,33 +37,136 @@ on {{manubot.date}}.
...
+[The Version 1.0 Deep Review Authors](#version-1.0-authors)
 
-<sup>♠</sup> --- Author order for version 2.0 is currently arbitrary.<br>
+<sup>♠</sup> --- Author order for version 2.0 is currently randomized [with each new build](https://github.com/greenelab/deep-review/pull/997).<br>
 <sup>†</sup> --- To whom correspondence should be addressed: greenescientist@gmail.com (C.S.G.)
 
+<small>
+
+{% for affiliation in manubot.affiliations %}
+2.{{affiliation.affiliation_number}}. {{affiliation.affiliation}}
```
dhimmel: I think jinja was stripping the trailing spaces. this should fix
```suggestion
2.{{affiliation.affiliation_number}}. {{affiliation.affiliation}}<br>
```

agitter:
What do we want to happen for an author who contributes to v1 and v2 and has a change in competing interests? Do we keep the old date and old COI in the v1 table but use the newer date and current COI in the v2 table?

cgreene:
@agitter : Not ideal, but that was the only thing I could come up with, unless we added a display true/false flag and kept the old author metadata.

agitter:
Yeah, I don't have a great way to improve upon the current approach.  When version 2.0 is close to finished, we could always manually update entries in the version 1.0 COI table.

PR #1028, opened 2021-12-18
Revise front matter content update note

agitter:
I'm revising the front matter language to indicate there is updated version 2.0 content but be less specific about the timing of the update.


agitter:
@cgreene does this updated phrasing in the deep review header look good to you?

cgreene (approved):
Yea - this looks good

